PR #55, opened 2019-01-04
BHoM_UI - Provide support for menus other than WinForm and WPF

### LINKED ISSUE #54: BHoM_UI: Provide support for menus other than WinForm and WPF
See comment from @awakeman in https://github.com/BHoM/Excel_Toolkit/issues/45#issuecomment-451188351 for reference

adecler:
Fixes #54.

Should enable contextual menu support in Excel that apparently doesn't expose WinForm or WPF directly.

No change needed in Grasshopper or Dynamo. I have kept the methods supporting directly WinForm and WPF since they will be the most common ones anyway so the user still gets a type strong interface for the general cases.

The overall architecture is admittedly a bit convoluted but this is necessary to keep type strong support all the way (nicer for things like OrganiseItems methods) while having a asynchronous discovery of the types for components items (needs to be done ahead of time to prevent lag on opening context menu) and menu type (only known when actually opening the menu).  I'll eventually review it when I have more time but this is doing the job for now.

@epignatelli , @ZiolkowskiJakub , I have added you as reviewers just to make sure this doesn't impact things on your side (it shouldn't).

### DIFF
--- a/BHoM_UI/Templates/Selector.cs
+++ b/BHoM_UI/Templates/Selector.cs
@@ -71,26 +71,36 @@ namespace BH.UI.Templates
 
         public void AddToMenu(ToolStripDropDown menu)
         {
-            if (m_Menu_WinForm == null)
-            {
-                m_Menu_WinForm = new SelectorMenu_WinForm<T>(m_ItemListStore[m_Key], m_ItemTreeStore[m_Key]);
-                m_Menu_WinForm.ItemSelected += M_Menu_ItemSelected;
-            }
+            if (m_SelectorMenu == null)
+                SetSelectorMenu(new SelectorMenu_WinForm<T>(m_ItemListStore[m_Key], m_ItemTreeStore[m_Key]));
 
-            m_Menu_WinForm.FillMenu(menu);
+            m_SelectorMenu.FillMenu(menu);
         }
 
         /*************************************/
 
         public void AddToMenu(System.Windows.Controls.ContextMenu menu)
         {
-            if (m_Menu_Wpf == null)
-            {
-                m_Menu_Wpf = new SelectorMenu_Wpf<T>(m_ItemListStore[m_Key], m_ItemTreeStore[m_Key]);
-                m_Menu_Wpf.ItemSelected += M_Menu_ItemSelected;
-            }
+            if (m_SelectorMenu == null)
+                SetSelectorMenu(new SelectorMenu_Wpf<T>(m_ItemListStore[m_Key], m_ItemTreeStore[m_Key]));
 
-            m_Menu_Wpf.FillMenu(menu);
+            m_SelectorMenu.FillMenu(menu);
+        }
+
+        /*************************************/
+
+        public void AddToMenu(object menu)
+        {
+            if (m_SelectorMenu != null)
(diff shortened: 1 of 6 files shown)


awakeman:
At a cursory glance this looks like this should do the trick, will use it with Excel and check that there's nothing missing for that case but I don't foresee anything.

Thanks.

BHoM_UI/Templates/Selector.cs
```diff
@@ -71,26 +71,36 @@ public Selector(IEnumerable<T> possibleItems, string key)
...
-            m_Menu_Wpf.FillMenu(menu);
+        public void AddToMenu(object menu)
+        {
+            if (m_SelectorMenu != null)
+                m_SelectorMenu.FillMenu(menu);
+        }
+
+        /*************************************/
+
+        public void SetSelectorMenu<M>(SelectorMenu<T, M> selectorMenu)
```
awakeman: The problem with this is two-fold

1. outside of the context of `Selector<T>` or `Caller.SetPossibleItems<T>` we can't (statically) know what T is
2. outside of the context of `Selector<T>` we don't have access to the private members that make up the `SelectorMenu<T, M>` constructor parameters

The former could be fixed by a factory method that takes a `Selector<T>` as dynamic and returns something specialised on T, but then we still don't have the `List<SearchItem>, Tree<T>` to pass into the constructor.

My thinking is that we could have a factory method inside `Selector` (and declared in `ISelector`) that takes a `Type` at runtime (which has to be an incomplete type that is a subclass of `SelectorMenu<T, M>` partially specialised on `M` but not `T` that has a constructor that takes the arguments `List<SearchItem>, Tree<T>`) and invokes that constructor dynamically, then. But I really don't like that, it seems very inelegant.

The other idea I've had is to declare a `SetItems(List<SearchItem>, Tree<T>)` in `ISelectorMenu<T>` that will be called by `Selector<T>.SetSelectorMenu<M>()`, removing the need to construct a `SelectorMenu<T,M>` with those values and the library that creates the `SelectorMenu<T, M>` can implement a factory that takes a `Selector<T>`; creates its `SelectorMenu<T,M>` implementation, specialised on `T` and `M`; and calls `Selector<T>.SetSelectorMenu<M>()` on its parameter.

Any other ideas of how to handle this statically?

adecler:
What about this?
This way, you can have a new SelectorMenu with an empty constructor and don't need to worry about the item list. You still need to pass the Caller.Selector as dynamic to another function but it should unblock you for now.
I am still not super happy with the whole infrastructure (we are trying to lower the threshold to entry, not raise it) so I will revisit it properly when I have the time but I hope this does the job for you in the meantime.

awakeman:
Yeah, this is exactly the type of thing i was thinking (I had made a very similar local change myself and was planning on pushing it to this branch anyway but didn't get there. 

> I am still not super happy with the whole infrastructure (we are trying to lower the threshold to entry, not raise it) so I will revisit it properly when I have the time but I hope this does the job for you in the meantime.

I second this, it still doesn't feel very elegant nor simple, but it does do the job. Hopefully UIs are not the kind of thing people want to be adding regularly so the accessibility is less critical than Engines, oMs and Adapters.

Thanks.

PR #59, opened 2019-01-10
BHoM_UI - Issue 58 - provide the location of the new component to be created when using gobal menu

### LINKED ISSUE #58: BHoM_UI: guarantee public access to WinForm m_SearchTextBox
https://github.com/BHoM/Grasshopper_Toolkit/issues/311 requires access to the location of either the searchbox or the any other form element.

@adecler does this create problem from the general UI side?

adecler:
Fixes #58.

@epignatelli , let me know if that does it for you. You will still probably have to translate that into coordinates inside the canvas but it should otherwise do the trick.

### DIFF
--- a/BHoM_UI/BHoM_UI.csproj
+++ b/BHoM_UI/BHoM_UI.csproj
@@ -52,6 +52,10 @@
     <Reference Include="Geometry_Engine">
       <HintPath>..\..\BHoM_Engine\Build\Geometry_Engine.dll</HintPath>
     </Reference>
+    <Reference Include="Geometry_oM, Version=2.0.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\BHoM\Build\Geometry_oM.dll</HintPath>
+    </Reference>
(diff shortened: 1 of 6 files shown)


al-fisher:
@adecler  FYI confirmed the referencing of specific versioning of .dlls is behaving as expected - it does seem to be. Same effect as so whether .csproj references in the form

``` C#
    <Reference Include="Geometry_oM, Version=2.0.0.0, Culture=neutral, processorArchitecture=MSIL">
      <SpecificVersion>False</SpecificVersion>
      <HintPath>..\..\BHoM\Build\Geometry_oM.dll</HintPath>
    </Reference>
```

or 

``` C#
    <Reference Include="Geometry_oM">
      <HintPath>..\..\BHoM\Build\Geometry_oM.dll</HintPath>
    </Reference>
```

We have a mixture of both, I pressume as a result of manual .csproj modifcations. 
We should document best practice as well as add to future compliance checks.

adecler:
@al-fisher, I much prefer the second option but I think VS uses the first one by default so you have to manually change it. I wonder if there is a way for make sure VS does it correctly by default.

PR #61, opened 2019-01-14
Fixes #60 - setting inputs with specific type

### LINKED ISSUE #60: BHoM_UI: set caller InputParams with specific type

epignatelli:
Fixes #60 

Adding a `List<types>` input to the `SetInputs` method, required to pass specific types from the UI into the `CustomData`, see https://github.com/BHoM/Grasshopper_Toolkit/pull/310 for more.

The parameter has a default value so that we keep the same API and do not break things.
Also, if there is any issue with that it falls back to setting the inputs without their type.

You can use the file in the Grasshopper pr to test this.

### DIFF
--- a/BHoM_UI/Components/oM/CreateCustom.cs
+++ b/BHoM_UI/Components/oM/CreateCustom.cs
@@ -79,9 +79,17 @@ namespace BH.UI.Components
         /**** Public Methods              ****/
         /*************************************/
 
-        public void SetInputs(List<string> names)
+        public void SetInputs(List<string> names, List<Type> types = null)
         {
-            InputParams = names.Select(x => GetParam(x)).ToList();
+            if (names.Count != types.Count)
+            {
+                Engine.Reflection.Compute.RecordWarning("The list length for names and types does not match. Inputs are set, but <types> variable will be ignored.");
+                InputParams = names.Select(x => GetParam(x)).ToList();
+            }
+            else
+            {
+                InputParams = names.Select((x, i) => GetParam(names[i], types[i])).ToList();
+            }
 
             CompileInputGetters();
             CompileOutputSetters();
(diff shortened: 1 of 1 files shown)


BHoM_UI/Components/oM/CreateCustom.cs
```diff
@@ -79,9 +79,17 @@ public CreateCustomCaller() : base()
         /**** Public Methods              ****/
         /*************************************/
 
-        public void SetInputs(List<string> names)
+        public void SetInputs(List<string> names, List<Type> types = null)
         {
-            InputParams = names.Select(x => GetParam(x)).ToList();
+            if (names.Count != types.Count)
```
adecler: if type is null (default value of the parameter), this will crash.

BHoM_UI/Components/oM/CreateCustom.cs
```diff
@@ -79,9 +79,17 @@ public CreateCustomCaller() : base()
...
+            {
+                Engine.Reflection.Compute.RecordWarning("The list length for names and types does not match. Inputs are set, but <types> variable will be ignored.");
+                InputParams = names.Select(x => GetParam(x)).ToList();
+            }
+            else
+            {
+                InputParams = names.Select((x, i) => GetParam(names[i], types[i])).ToList();
```
adecler: Not the biggest deal but there is a specific Linq method for that:
`InputParams = names.Zip(types, (name, type) => GetParam(name, type)).ToList();`

BHoM_UI/Components/oM/CreateCustom.cs
epignatelli: Definitely, thanks for spotting it!

BHoM_UI/Components/oM/CreateCustom.cs
epignatelli: Nice! Thanks, I didn't know that!

PR #69, opened 2019-01-24
BHoM_UI - Issue 68 - Explode fails for null in custom data or dictionary

### LINKED ISSUE #68: BHoM_UI: Explode fails for null in custom data or dictionary
If the custom data contains any null values, the explode component crashes down.

This currently causes a full Rhino crash if done from inside grasshopper.

Test file:

[ExplodeNullFail.zip](https://github.com/BHoM/BHoM_UI/files/2791529/ExplodeNullFail.zip)

epignatelli:

Closes #68

### Test files

[ExplodeNullFail.zip](https://github.com/BHoM/BHoM_UI/files/2791529/ExplodeNullFail.zip) (Thanks @IsakNaslundBh)

### DIFF
--- a/BHoM_UI/Components/Engine/Explode.cs
+++ b/BHoM_UI/Components/Engine/Explode.cs
@@ -183,7 +183,7 @@ namespace BH.UI.Components
                         {
                             foreach (string key in dic.Keys.OfType<string>())
                             {
-                                if (!properties.ContainsKey(key))
+                                if (key != null & dic[key] != null &!properties.ContainsKey(key))
                                     properties[key] = dic[key].GetType();
                             }
                         }
@@ -198,7 +198,7 @@ namespace BH.UI.Components
                 {
                     foreach (KeyValuePair<string, object> prop in group.Cast<BHoMObject>().SelectMany(x => x.CustomData).Distinct())
                     {
-                        if (!properties.ContainsKey(prop.Key))
+                        if (!properties.ContainsKey(prop.Key) & prop.Value != null)
                             properties[prop.Key] = prop.Value.GetType();
                     }
                 }


rwemay:
Hi guys, trying to build this as explode is still not working for me in GH - hasn't done for some time now. Should this PR reference the BHoM_Engine PR branch that it seems to require to build? 

https://github.com/BHoM/BHoM_Engine/pull/781

image

adecler:
Hi @rwemay , 
Yes, same problem for me. I had to switch to the branch in the engine for this to compile.
As for the explode component not working, you should check https://github.com/BHoM/Grasshopper_Toolkit/pull/321.  A few minor changes needed before it can be merged but it is working great for me otherwise.

adecler:
@epignatelli , I think you went a bit overboard with your last commit. The intent is good but this is a difficult thing to get right and I would rather have you sorting your pending requests instead of going in that direction. If you solve my last comment by doing something simple like finding the first non-null element or something, that would be fine. If you want to be a bit more fancy, you can collect a `HasSet` of all the types registered with that property and take the first one if the only one in the list or object otherwise. But keep your code simple. The change should not be more than a few line of code. Your last commit is a bit scary for something that trivial :-P.

epignatelli:
I probably haven't caught the issue you were talking about then.
What I was trying to fix is:
- We have multiple objects (same or different type), having the same property name `X`
- For the first object in the group, the value of `X` is `null`.
- The type of `X` is set to `object` for both the objects because the first one is null

In the `foreach` loop I am not yet aware of the of the object I haven't iterated on yet. Were you suggesting to iterate again on the groups within the scope?

Is that the problem, right?

If it is, and you're concerned about getting the common parent, I can just replace that.

epignatelli:
Apologies guys for the undeclared dependency. I just removed it, you can be on the BHoM_Engine master branch now.

al-fisher:
hey @IsakNaslundBh can you re-look at this now in light of @epignatelli latest commits to see if the requested changes have been address? Thanks

al-fisher:
Brilliant- thanks @adecler! Well done @epignatelli

PR #75, opened 2019-01-31
BHoM_UI-Issue64-GetPropertyHandlingCollections

### LINKED ISSUE #64: BHoM_UI: GetPropertyComponent does not unravel collections
It records the `OutputType()` of the method as `DataType` of output parameter of the component.
`PropertyValue` returns a type `object`.

This prevent collection types being set correctly and the `DataAccessor` compiles according to the wrong depth.

epignatelli:
Getting runtime type of the collected property value rather than declared type (GetProperty returns `object`)

Closes #64
The component was setting the output type base on the return type of the `GetPropertyValue` method, which is `object`. This creates problem when the runtime type is a `System.Collection`, preventing the component from unravelling it.
We're now getting the runtime type of the output. If `null` we're bypassing it and setting the declare type.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/Es3A0aBdb11DibWCl17DCPwBgb0wmqEE9acA4vi50b_OLQ?e=Wnx240

#### Added:
- `public bool GetPropertyCaller.SetOutputTypes(object result)`
- `public override object GetPropertyCaller.Run(object[] inputs)`

### DIFF
--- a/BHoM_UI/Components/Engine/GetProperty.cs
+++ b/BHoM_UI/Components/Engine/GetProperty.cs
@@ -21,15 +21,9 @@
  */
 
 using BH.oM.Reflection.Attributes;
-using BH.oM.UI;
 using BH.UI.Templates;
 using System;
-using System.Collections.Generic;
 using System.ComponentModel;
-using System.Linq;
-using System.Reflection;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace BH.UI.Components
 {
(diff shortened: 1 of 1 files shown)


epignatelli:
Thanks all! It would be good to have a green light from @adecler for the general architecture, but I'd rather merge this as quickly as possible.

@al-fisher I am happy if this is merged as it doesn't create any new issue!

al-fisher:
Fantastic thanks all - works and all makes sense to me. I am happy to merge.
@adecler - keeping you @'d so if you want to peep after  🎆 New Year 🎆 and want to raise an architecture or compliance tweak feel free! 😃  Cheers

PR #82, opened 2019-03-29
Issue74-ExplodeCustomBaseProps

### LINKED ISSUE #74: BHoM_UI: Exploding custrom objects should return the base properties
Currently, exploding a custom object only gives the values in the custom data.

The explode also needs to return the base properties from the BHoMObject:

-Name
-BHoM_Guid
-Tags

image

Will be increasingly important to be able to stitch objects back together after https://github.com/BHoM/BHoM_Engine/pull/782 has been merged

epignatelli:

Closes #74 

For `CustomObject`, this PR adds the base `BHoMObject` properties to the list of fields captured in the explode.
It adds them only if they are defined.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/Egdx6sfWpplHnkdPH1R9EscBKtUpuoff95_8dXXabETf8g?e=avojRp

### DIFF
--- a/BHoM_UI/Components/Engine/Explode.cs
+++ b/BHoM_UI/Components/Engine/Explode.cs
@@ -200,13 +200,20 @@ namespace BH.UI.Components
                 }
                 else if (group.Key == typeof(CustomObject))
                 {
-                    foreach (KeyValuePair<string, object> prop in group.Cast<BHoMObject>().SelectMany(x => x.CustomData).Distinct())
+                    IEnumerable<CustomObject> objs = group.Cast<CustomObject>();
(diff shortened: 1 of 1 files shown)


adecler:
Works great for me.

ExplodeBHoMProperties

My only comment would be that we should have consistency between the way explode behaves with CustomObjects and regular BHoM objects. At the moment, exploding a BHoM object will always output the Names, Tags and CustomData regardless of them being empty or not. Either behaviour is fine with me but it should be consistent accross.

image

epignatelli:
Thanks @adecler ! I removed the check for default values for the `CustomObject`, I'd rather always having them to reduce the number of modifications to the UI component.

PR #84, opened 2019-04-12
#Fixes #83

### LINKED ISSUE #83: BHoM_UI: Remove "pulls everything by default" in the pull method description

epignatelli:

Closes #83 

Removes the part of the description that explicitly says that a "Pull all" behaviour is triggered when no query is used, since it is true only for specific adapters (Socket).


epignatelli:
Thanks here as well @alelom 

Fine for this to be merged when possible

Issue #72, opened 2019-01-30
BHoM_UI: Set correct output for explode when values of different types are used for the same key


IsakNaslundBh:
When different types are fed in to the same key, the explode component would need to find a common denominator between the values set to the key.

For example, setting different types of geometry to the same key in a custom object.

See test file:

[DifferentTypeCustomData.zip](https://github.com/BHoM/BHoM_UI/files/2816284/DifferentTypeCustomData.zip)

Linked to what @epignatelli started working on here:

https://github.com/BHoM/BHoM_Engine/issues/780

adecler:
@IsakNaslundBh , can you provide a bit more information?

IsakNaslundBh:
Sorry, @adecler . I have updated the description.

PR #90, opened 2019-05-10
Different UIs concur to fill the BHoM/Assembly folder

### LINKED ISSUE #89: PostBuild_UI: Different UI builds concur to override the BHoM/Assemblies folder
As a result, you get only the latest UI you compiled. 
i.e. Excel removes Grasshopper files and Grasshopper removes Excel files

epignatelli:

Closes #89 

When compiling a UI, the BHoM/Assemblies folder gets cleared. This means that if you build two UIs, you will get only the second one, since the first one gets wiped out.

I am adding a check for specific files, but this is not scalable. 
For the moment it is sufficient though, as we won't add a new UI every day.
 

### Test files

No test file. To test, compile the Grasshopper_Toolkit first and then compile the Excel_Toolkit.
If the BHoM/Assemblies folder still contains the `.gha` file required by Grasshopper, the issue is considered fixed.

### DIFF
--- a/UI_PostBuild/Program.cs
+++ b/UI_PostBuild/Program.cs
@@ -130,9 +130,12 @@ namespace BHoM_UI
         private static void CleanDirectory(string folder)
         {
             DirectoryInfo di = new DirectoryInfo(folder);
-
+            List<string> skipThese = new List<string> { ".gha", ".xll", ".dna", ".Addin" };
             foreach (FileInfo file in di.GetFiles())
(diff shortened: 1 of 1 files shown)


epignatelli:
@IsakNaslundBh I fixed #72, but this is blocking. 
If this pr fixes the UI concurrency, can you approve?

FraserGreenroyd:
@adecler to review following BHoM call discussions :smile:

PR #95, opened 2019-06-12
BHoM_UI - Issue 81 - deprecated query not hidden

### LINKED ISSUE #81: BHoM_UI: Deprecated query method is showing up in global serach
For the global ctrl+b search, query methods modifying a query shows up 

image

The method can not be accessed via the modify component or the createQuery component. Only shows up for the global menu.

epignatelli:

Closes #81 

Two actions on this pr:
1. The global menu now shows methods that return an `IQuery` object just once. On the master branch, the method will appear twice: once with the `CreateQuery` icon, and once with the `CreateObject` icon.
2. The `CreateQuery` component now only exposes the methods from the `Create` class, as its nae would suggest. At the moment, any method that returns an `IQuery` (e.g. a `Modify`) will appear in the local menu.

@ZiolkowskiJakub I think the biggest effect of this pr is on the Revit_Toolkit, and also it creates a user case to see of the menu works.
Since we are using the `ReturnType` of the method to build the menu tree, all the Revit_Toolkit queries are now in the `DataManipulation` branch, and not in the Revit branch anymore. I don't find this very intuitive, but if we use the namespace of the `DeclaringType` as discriminator, we would have no way to know when an object is a `NonBHoMObject` (i.e. every `Create` method comes from the `Engine`).

### Test files

To test, compile the BHoM_UI and the Grasshopper_Toolkit; open the Grasshopper and check both the global menu (CTRL+SHIFT+B) and the `CreateQuery` local menu.

### DIFF
--- a/BHoM_UI/Components/Adapter/CreateQuery.cs
+++ b/BHoM_UI/Components/Adapter/CreateQuery.cs
@@ -57,7 +57,11 @@ namespace BH.UI.Components
         public CreateQueryCaller() : base()
         {
             Type queryType = typeof(BH.oM.DataManipulation.Queries.IQuery);
-            IEnumerable<MethodBase> methods  = BH.Engine.Reflection.Query.BHoMMethodList(true).Where(x => x.DeclaringType.Name == "Create" && queryType.IsAssignableFrom(x.ReturnType)).OrderBy(x => x.Name);
+            IEnumerable<MethodBase> methods  = BH.Engine.Reflection.Query.BHoMMethodList()
+                .Where(x => x.DeclaringType.Name == "Create"
+                && queryType.IsAssignableFrom(x.ReturnType)
+                && !x.IsDeprecated())
+                .OrderBy(x => x.Name);
             SetPossibleItems(methods);
         }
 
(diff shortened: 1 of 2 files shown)


BHoM_UI/Components/Adapter/CreateQuery.cs
```diff
@@ -57,7 +57,7 @@ public class CreateQueryCaller : MethodCaller
         public CreateQueryCaller() : base()
         {
             Type queryType = typeof(BH.oM.DataManipulation.Queries.IQuery);
-            IEnumerable<MethodBase> methods  = BH.Engine.Reflection.Query.BHoMMethodList().Where(x => queryType.IsAssignableFrom(x.ReturnType)).OrderBy(x => x.Name);
+            IEnumerable<MethodBase> methods  = BH.Engine.Reflection.Query.BHoMMethodList(true).Where(x => x.DeclaringType.Name == "Create" && queryType.IsAssignableFrom(x.ReturnType)).OrderBy(x => x.Name);
```
IsakNaslundBh: This does not build on my machine. `BH.Engine.Reflection.Query.BHoMMethodList(true)` does not take any argument. Guessing you need a engine branch for this as well?

epignatelli: Thanks @IsakNaslundBh, your guess is definitely right - I pushed a fixed.

epignatelli:
Thanks @alelom, the menu I was referring to is the one that pops when create a `CreateQuery` component and right click on that.

epignatelli:
Thanks very much @alelom - merging to allow solving https://github.com/BHoM/BHoM_UI/issues/96.

PR #99, opened 2019-06-26
BHoM_UI - Issue 89 - Include interface types into the CreateTypes component

### LINKED ISSUE #89: PostBuild_UI: Different UI builds concur to override the BHoM/Assemblies folder

### LINKED ISSUE #98: BHoM_UI: include interface types in CreateType component

epignatelli:
Depends on: https://github.com/BHoM/BHoM_Engine/pull/1062

 
Closes #98 

This change follows the on e in the dependent pull request in the `Reflection_Engine`.
The `CreateTypes` component now includes these interface types.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/Emt9HMps2LpCrd4zxetp_4MBOHYnPOmtva-yzsLO9BDuvA?e=pSFHiz

### DIFF
--- a/UI_Engine/Query/SearchItems.cs
+++ b/UI_Engine/Query/SearchItems.cs
@@ -107,6 +107,10 @@ namespace BH.Engine.UI
         {
             return Engine.Reflection.Query.BHoMTypeList()
                 .Concat(Engine.Reflection.Query.BHoMInterfaceList())
+                .Concat(new List<Type> { typeof(Type), typeof(Enum),
+                    typeof(object), typeof(bool), typeof(byte),
+                    typeof(char), typeof(string),
+                    typeof(float), typeof(double), typeof(decimal), typeof(short), typeof(int), typeof(long),})
                 .Where(x => !x.IsNotImplemented() && !x.IsDeprecated());
         }
 
--- a/UI_Engine/Query/SearchItems.cs
+++ b/UI_Engine/Query/SearchItems.cs
@@ -35,14 +35,14 @@ namespace BH.Engine.UI
         /**** Public Methods                            ****/
         /***************************************************/
 
-        public static IEnumerable<MethodInfo> EngineItems()
+        public static IEnumerable<MethodBase> EngineItems()
         {
             return Engine.Reflection.Query.BHoMMethodList().Where(x => !x.IsNotImplemented() && !x.IsDeprecated());
         }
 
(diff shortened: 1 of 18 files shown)


IsakNaslundBh:
To capture our conversation Eduardo:
- The interface types show up in the global search menu once for each `CreateType` component on the canvas the first time you initialise the global search menu
- The interface types does not show up in the global search menu in excel, but can be constructed using the inline formula @awakeman 
- The global search in dynamo does not seem to trigger for me, so cant check that, but I am able to create the interface type in dynamo.

epignatelli:
Thanks very much @IsakNaslundBh for catching that - I pushed a fix.

@ZiolkowskiJakub @awakeman FYI, now the global menu (CTRL + SHIFT + B) and the local menu (right click on component, or `=` with formula) return exactly the same items - this would be reflected in Dynamo and Excel as well.

UI_Engine/Query/SearchItems.cs
```diff
@@ -107,7 +107,6 @@ public static IEnumerable<Type> TypeItems()
         {
             return Engine.Reflection.Query.BHoMTypeList()
                 .Concat(Engine.Reflection.Query.BHoMInterfaceList())
-                .Concat(new List<Type> { typeof(object), typeof(Type), typeof(Enum), typeof(bool), typeof(byte), typeof(char), typeof(double), typeof(int), typeof(string) })
```
IsakNaslundBh: Not to sure why you are getting rid of these? Think we want them for generating dictionaries of specific types etc? I know I have used them at least a few times.

epignatelli: The main reason why I removed these is because the local menu is completely wrong if we include them. But I agree, that was bad choice, I'll try to fix the menu instead.

epignatelli: How do you feel with this? It looks alright to me
image

IsakNaslundBh: I would agree with that solution

epignatelli: So, this solution creates other problems, and since it is not due to solve it in this pr, I will leave it as it is in the master right now

PR #106, opened 2019-07-01
Add keyboard control to SearchMenu

### LINKED ISSUE #105: BHoM_UI: add keyboard control to GlobalSearch
Just a thought from some of my use of GlobalSearch recently, it is summoned with the keyboard, the keyboard is used to search but then you have to select with the mouse which is jarring and doesn't feel great from a UX perspective.

Should be simple to add some keybindings to move up and down in the list and select.

Suggested behaviour:

`Up`/`Down` - highlight a result in the list
`Enter` - Select highlighted result
`Esc` - Cancel (close the window)

Top result should be highlighted automatically.

awakeman:

Closes #105 

Currently only adds functionality to WinForms search menu (Excel and Grasshopper). Unable to test with Dynamo because I can't open the GlobalSearch on Dynamo to do so. Have the change on that side too that I think should work but I can't verify it.

### DIFF
--- a/BHoM_UI/Global/SearchMenu_WinForm.cs
+++ b/BHoM_UI/Global/SearchMenu_WinForm.cs
@@ -91,7 +91,8 @@ namespace BH.UI.Global
                     Size = new System.Drawing.Size(m_MinWidth, 1),
                     TabIndex = 1,
                 };
-                m_Popup.Controls.Add(m_SearchResultPanel);  
+                m_Popup.Controls.Add(m_SearchResultPanel);
+                m_Popup.KeyDown += M_Popup_KeyDown;
             }
 
             // Finish the popup form
@@ -113,7 +114,42 @@ namespace BH.UI.Global
 
             TextBox_TextChanged(null, null);
 
+
             return true;
+        }
+
+        private void M_Popup_KeyDown(object sender, KeyEventArgs e)
+        {
+            e.SuppressKeyPress = true;
+            switch (e.KeyData)
+            {
+                case Keys.Up:
+                    if (--m_selected < 0)
+                        m_selected = m_SearchResultPanel.Controls.Count - 1;
+                    break;
+                case Keys.Down:
+                    m_selected = (m_selected + 1) % m_SearchResultPanel.Controls.Count;
+                    break;
+                case Keys.Enter:
+                    List<SearchItem> hits = PossibleItems.Hits(m_SearchTextBox.Text);
+                    if (m_selected < hits.Count)
+                    {
+                        m_Popup.Hide();
+                        NotifySelection(hits[m_selected]);
+                    }
+                    return;
+                case Keys.Escape:
+                    m_Popup.Hide();
+                    return;
+                default:
+                    e.SuppressKeyPress = false;
+                    return;
+            }
+            for (int i = 0; i < m_SearchResultPanel.Controls.Count; i++)
+            {
+                Control row = m_SearchResultPanel.Controls[i];
+                row.BackColor = (i == m_selected) ? System.Drawing.SystemColors.MenuHighlight : Color.Transparent;
+            }
         }
 
 
@@ -130,6 +166,7 @@ namespace BH.UI.Global
 
         private void TextBox_TextChanged(object sender, EventArgs e)
         {
+            m_selected = 0;
             List<SearchItem> hits = PossibleItems.Hits(m_SearchTextBox.Text);
 
             int yPos = 0;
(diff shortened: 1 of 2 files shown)


awakeman:
Note: I have pushed the change for WPF but haven't actually tested from within dynamo yet. So it may not work as well.

FraserGreenroyd (dismissed):
Have tested solely in Grasshopper, so will need someone to review in Excel/Dynamo, however, in GH this is working very nicely - brings back some nice functionality from last year I think which certainly makes the global search easier to use for people like me who are keyboard focused.

Approved from a Grasshopper perspective - nice work! :smile:

michaldengusiak (dismissed):
absolutely amazing!!! thanks

epignatelli (changes requested):
That's a really great addition - thanks, @awakeman!! 🚀 

Grasshopper is working great. I am afraid Dynamo side does not work yet though, I can't go up and down the menu.
I want to loop @ZiolkowskiJakub in the conversation to make sure that that side is fully functional.
Have you tested Excel?

Code-wise, the methods for both WinForms and Wpf have the same interface. 
Is there any drawback in promoting it up at the `SearchMenu` object level?

Super minor - what about a little opacity to the highlight color? That's pretty vivid right now.

michaldengusiak:
@epignatelli 
I tested in excel and works perfect
image

epignatelli:
> @epignatelli
> I tested in excel and works perfect
> image

Thanks @michaldengusiak - waiting for @ZiolkowskiJakub  to test Dynamo and @awakeman for the changes.

michaldengusiak:
@epignatelli dynamo got  `Ctrl+Shift+b` disable  
but when using search, it works fine...   **maybe we can enable** in Dynamo  `Ctrl+Shift+b` again?
image

awakeman:
@epignatelli the main issue with promoting it up is that the interface may look the same but the methods actually take different types, (the `KeyEventArgs` are in different namespaces and the bodies couldn't necessarily be generalised to a useful degree)

All the other event handling stuff between the two is done at the subclass level too so makes sense to keep it that way to me.

epignatelli:
> All the other event handling stuff between the two is done at the subclass level too so makes sense to keep it that way to me.

That's ok like for me - let me now when you plan to do the other changes.

awakeman:
Also I opted for using the system colours because then it will fit with whatever the user has selected for their highlight colour. Most importantly meaning someone who uses a scheme specifically for accessibility reasons (e.g. high contrast) will not have those settings overridden by our aesthetic preferences.

epignatelli:
Got it, Dynamo still doesn't work though.

awakeman:
@epignatelli could you show me what dynamo's one looks like on master? i can't get it to work on either.

epignatelli:
Hey, sorry – I missed the skype,

No, I don’t think your pr broke something. I am able to get the global menu like I was on the master.
You need to instantiate the first component on the canvas for the menu to start listening to key-down events.

This is how it looks like to me, no selection happening via keyboard:
 dasads

ZiolkowskiJakub (changes requested):
Hi,

The same comment as @epignatelli about arrow keys. It does not work in Dynamo.

I also cannot access DropDown Menu when running Dynamo directly from Revit. It works fine for Sandbox mode. I believe the difference is that when Dynamo open from Revit it does not use sockets but not sure if this cause the issue (@IsakNaslundBh should know more about this part).

@michaldengusiak : I believe your issue with text not being shown is related to scaling mode of component. I have Display Settings scale set to 100% and it looks good. @awakeman what do you think? Just in case are you able to change scaling option of component?

Thanks!

awakeman:
Should be fixed now. The dodgy scaling bugs and it not working when running inside Revit appear to be separate and existing bugs (they are there when using the `master` branch). But now the up and down keys work.

michaldengusiak (approved):
correct, tested in Dynamo Sandbox and can scroll down 
image
 however when I type Panel still got this...
image

will check  scaling I had 120%, now changed to 100% and works perfectly
image

tested in Revit Dynamo and it does not work for me `Ctrl+Shift+b`

epignatelli (approved):
LGTM! Thanks @awakeman, that's much better now.

I broadly checked dynamo, but I'll wait for @ZiolkowskiJakub to do a formal check.

ZiolkowskiJakub (approved):
Hi,

Works good now.

As discussed:
Dynamo Sandbox works fine but have trouble with run it directly in Revit. However, this would be Dynamo_Toolkit issue.

awakeman:
Thanks all

PR #116, opened 2019-07-05
BHoM_UI - ISSUE 394 - Add hints and tree access to property value of SetProperty

### LINKED ISSUE #91: BHoM_UI: In deserialisationm, if caller doesn't find a suitable method, InputParams and OutputParamas are not set
If a component fails to deserialise, the `SelectedItem` is correctly set to null, but the `InputParams` and `OutputParams` are incorrectly set to an empty list.

This makes an "unmorphed" component equal to a "morphed" one whose serialisation fails (specifically, whose backend method is not found).
i.e. the two `CreateObject` components below are, BHoM-wise, equal. All the extra information we see, come from the Grasshopper part of the serialisation.

All of this cancels any possibility to separate out the badly-deserialised component from  new, unmorphed compoenents. Ultimately, it does not allow https://github.com/BHoM/Grasshopper_Toolkit/issues/220

To solve this, in the `catch` snippet below, we should set the `InputParams` and the `OutputParams` .
image

https://github.com/BHoM/BHoM_UI/blob/bcaca5ffa67e64d10d206f77532bbe0184488d9a/BHoM_UI/Templates/Caller.cs#L183-L201

thanks @alelom

### LINKED ISSUE #112: BHoM_UI: Align to left
@epignatelli as we discuss `align to left` to avoid text jumping on screen depends on methods length

image
image

epignatelli:

This is mainly required to support the fixes in: https://github.com/BHoM/Grasshopper_Toolkit/pull/385, but also:
Closes #112
Closes #91 

I left some messages on the code itself to make it easier for the reviewers to track the changes.
To summarise:
- The `SetProperty` requires to serialise features of the `InputParams` as well (e.g. <list input>). The same serialisation is used in the `CreateCustom`. This pr then generalises this serialisation into the parent `Caller` class. It is now serialised into a `CustomObject` with: `SelectedItem`, `InputParams` and `OutputParams`. 
- `CreateCustom.AddInput`, `CreateCustom.RemoveInput` and `CreateCustom.UpdateInput` are now used by the `SetProperty` component as well. They are thus migrated to the parent `Caller`.
- The `CreateCustom.Read` and `CreateCustom.Write` are used also by the `SetProperty`. Also, due to #91, compoenents read with a `SelectedItem == null` could be distinguished by unmorphed ones. Check the body of #91 for more. Serialisation has been then migrated to the general `Caller`.
- `CreateCustom.GetParam` overloads have been replaced by:
  - `Engine.UI.Convert.ToBHoM(PropertyInfo),
  - `Engine.UI.Create.ParamInfo(string, Type, ParamKind)

### Test files

@peterjamesnugent , @michaldengusiak , @IsakNaslundBh  I included you in the reviewers to ask you if you can check that serialisation is working for your old scripts as well. Last thing I would like is to improve serialisation and break the old one.
@adecler and @alelom could you check instead if the code structure makes sense and there are no bugs we can see?


BHoM_UI/Components/Engine/SetProperty.cs
```diff
@@ -56,27 +50,19 @@ public class SetPropertyCaller : MethodCaller
         /**** Constructors                ****/
         /*************************************/
 
-        public SetPropertyCaller() : base(typeof(SetPropertyCaller).GetMethod("SetProperty")) { }
+        public SetPropertyCaller() : base(typeof(Engine.Reflection.Modify).GetMethod("PropertyValue")) { }
```
epignatelli: I don't see the need to wrap the component around a new method that calls the backend one. 
The name is solved by overriding the `Name` property.

BHoM_UI/Components/oM/CreateCustom.cs
```diff
@@ -78,64 +79,21 @@ public CreateCustomCaller() : base()
         /**** Public Methods              ****/
         /*************************************/
 
-        public void SetInputs(List<string> names, List<Type> types = null)
+        public void SetInputParams(List<string> names, List<Type> types = null)
```
epignatelli: Aligns to `MethodCaller.SetInputParams`

BHoM_UI/Components/oM/CreateCustom.cs
```diff
@@ -78,64 +79,21 @@ public CreateCustomCaller() : base()
...
-                InputParams = names.Zip(types, (name, type) => GetParam(name, type)).ToList();
-            }
-
-            CompileInputGetters();
-            CompileOutputSetters();
-        }
-
-        /*************************************/
-
-        public bool AddInput(int index, string name, Type type)
```
epignatelli: Inputs handling migrates to the parent `Caller` as it is used by all components

BHoM_UI/Components/oM/CreateCustom.cs
```diff
@@ -168,103 +126,23 @@ public override bool SetItem(object item)
             {
                 Name = ForcedType.Name;
                 Description = ForcedType.Description();
-                InputParams.AddRange(ForcedType.GetProperties().Select(x => GetParam(x)).ToList());
+                InputParams.AddRange(ForcedType.GetProperties().Select(x => x.ToBHoM()).ToList());
```
epignatelli: GetParam is replaced by a `Convert` method for compliance.

BHoM_UI/Components/oM/CreateCustom.cs
```diff
@@ -168,103 +126,23 @@ public override bool SetItem(object item)
...
             }
             return true;
         }
 
         /*************************************/
 
-        public override string Write()
```
epignatelli: Now all components serialise also `InputParams` and `OutputParams` - moved to the general `Caller`.

BHoM_UI/Global/SearchMenu_WinForm.cs
```diff
@@ -75,7 +75,7 @@ public override bool SetParent(object parent)
                     Location = new System.Drawing.Point(0, 0),
                     Size = new System.Drawing.Size(m_MinWidth, 22),
                     TabIndex = 0,
-                    TextAlign = System.Windows.Forms.HorizontalAlignment.Center,
+                    TextAlign = System.Windows.Forms.HorizontalAlignment.Left,
```
epignatelli: This fixes #112

BHoM_UI/Templates/Caller.cs
```diff
@@ -167,13 +205,15 @@ public virtual string Write()
         {
             try
             {
-                if (SelectedItem == null)
-                    return "";
-                else
-                    return SelectedItem.ToJson();
+                CustomObject component = new CustomObject();
```
epignatelli: We now serialise also `InputParams` and `OutputParams`, since the user may have made some changes to the component that we want to keep track of.
Also, if the backend method does not exist anymore, the component does not reset to no inputs and no outputs.
This will help solve #91, to add the feature in https://github.com/BHoM/Grasshopper_Toolkit/issues/220

BHoM_UI/Templates/Caller.cs
```diff
@@ -232,7 +298,7 @@ protected virtual bool PushOutputs(object result)
                 else if (m_CompiledSetters.Count > 0)
                 {
                     for (int i = 0; i < m_CompiledSetters.Count; i++)
-                        m_CompiledSetters[i](DataAccessor, BH.Engine.Reflection.Query.Item(result as dynamic, i));
+                        m_CompiledSetters[i](DataAccessor, BH.Engine.Reflection.Query.IItem(result, i));
```
epignatelli: compliance

BHoM_UI/Templates/MethodCaller.cs
```diff
@@ -124,7 +124,7 @@ public override object Run(object[] inputs)
             {
                 return m_CompiledFunc(inputs);
             }
-            else if (inputs.Length <= 0)
+            else if (InputParams.Count <= 0)
```
epignatelli: bug

epignatelli:
Thanks very much @michaldengusiak ! 
I think excel needs fix, I'll dig into that.

epignatelli:
@michaldengusiak excel should be fixed in https://github.com/BHoM/Excel_Toolkit/pull/120

epignatelli:
> Noticed that the text in SearchMenu isn't updated to reflect the change in some way.
> 
> https://github.com/BHoM/BHoM_UI/blob/a1da53e11b9264b622e8cc301e8e87beb8128dc4/BHoM_UI/Global/SearchMenu.cs#L99-L100

Not sure I got that. Do you mean it should be `BH.Engine.Reflection.Modify.PropertyValue`?
If so, the current string is to differentiate the component from the back-end method in the global menu.

awakeman:
@epignatelli yeah, I get that they are to make it different but it's kinda odd that they claim to be a method that no longer exists. not sure what the right thing to do is. Maybe just leave it so it doesn't change name in the globalsearch.

epignatelli:
Yeah, I agree, I would leave it as it is right now. It should only affect the `string` distance in the search - if that works now, I feel it's ok.

epignatelli:
Thanks @awakeman, this is just waiting for https://github.com/BHoM/Grasshopper_Toolkit/pull/385 and https://github.com/BHoM/Dynamo_Toolkit/pull/158 to be reviewed

epignatelli:
All the dependent pull requests have been approved. 
Thanks all! We can now move the target to merge all of them on Monday.

PR #129, opened 2019-09-06
Removed event

### LINKED ISSUE #111: BHoM_UI: Update names and descriptions for GetProperty and SetProperty
Update input- and output names and add descriptions for component to make more intuitive

image

### LINKED ISSUE #127: BHoM sub-menu are disappearing to the next screen when method text is too long
#### Description:
See title and this:

image

#### How to replicate:
Add a `CreateObject` component to the canvas and, in the contextual menu, go to `Climate.EnvironmentCondition`

#### Expected behaviour:
The Menu should stay on the same screen

I would recommend simply limiting the size of the string representing the method to a given value. Will have to figure what a good value is though. We have 3 places where the text need to be truncated: Ctrl-Shift-B menu, search in the component menu and component menu itself. If there are multiple places where `ToText()` is called, refactor so that it is done inside a method that also limit the size.

alelom:

Closes #127
Closes #111

### DIFF
--- a/BHoM_UI/BHoM_UI.csproj
+++ b/BHoM_UI/BHoM_UI.csproj
@@ -182,7 +182,8 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
-    <PostBuildEvent>call "$(TargetDir)UI_PostBuild.exe" ..\..\ "C:\Users\$(Username)\AppData\Roaming\BHoM\Assemblies"</PostBuildEvent>
+    <PostBuildEvent>
+    </PostBuildEvent>
   </PropertyGroup>
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.


al-fisher:
Hi @alelom - as just discussed, proposing not removing the post build from BHoM_UI. Rather removing the calling of UI_PostBuild.exe from the individual GH, DYN, Excel. To avoid duplication of responsibilities. Individual implementation of UI's with just be responsible for any post builds needed for there own specific requirements.

epignatelli:
@al-fisher I was fought at the beginning as well.

My only worry about centralising the events is that, the BHoM_UI must always be compiled, in any case, even if there are no changes to it.

It may not be the case when, I am modifying the BHoM_Engine and the Grasshopper_Toolkit at the same time (for a related issue), but no API changes are relevant to the BHoM_UI.

I think it is a more common case than building all the three UIs one after the other.
So, in general, the idea behind this was to assign the events to the latest tip down the chain.

Although, I fine also doing it the other way around!

BHoM_UI/Components/Engine/GetProperty.cs
```diff
@@ -66,8 +66,10 @@ public override object Run(object[] inputs)
         /**** Public Method               ****/
         /*************************************/
 
-        [Description("Get the value of a property with a given name from an object")]
-        [Output("Value of the property")]
+        [Description("Get the value of a property or of a list of properties from an object.")]
+        [Input("obj", "Object to extract the property value from. Can be a list of objects.")]
+        [Input("propName", "Name of the property to extract. Can be a list of names. \nTo know what property names you can specify, try connecting your object to a BHoM `Explode` component.")]
```
epignatelli: Since we are specifying what the string can be, can we also explicitly say that you can sue the `.` operator?
e.g. for a `BH.oM.Structure.Bar`, you can use `StartNode.Position` to get the `BH.oM.Geometry.Point`

PR #147, opened 2019-10-23
CreateObject component to include default constructor for each BHoM object type

### LINKED ISSUE #124: BHoM_UI : Reflect Defining Property Create Methods into CreateObject menu
Include reflected create method, effectively exposing object initialiser capability for any combination of defining properties.
This has previously been exposed as part of Custom Object - needs to be migrated into  CreateObject menu.

adecler:
### NOTE: Depends on  
https://github.com/BHoM/BHoM_Engine/pull/1276

   
Closes #124

### Test files
Just pop the new component on the canvas and play with it.

### Additional comments
- This is just a first version of the component for testing. There are still quite a few things to sort out properly (such as the speed issue of not relying on existing methods). But should be good enough for you to try.
- To make it easier to compare behaviours, I have created a completely new component instead of replacing the existing one. This is obviously temporary and will be fixed before merging.
- This component will replace both the `CreateObject` and the `CreateCustom` components so bear that in mind when testing.
- I have only done the Grassopper UI PR for testing purposes. Will do the other UIs once we settle on the code and are ready to merge this.


epignatelli:
Hi @adecler ! I added the `status:WIP` label to the pr as it is a draft. I was not sure which one the two was the truth (no label or draft state). Let me know if I can review it, in case!

adecler:
Thanks @epignatelli , we are starting to have a bit too many labels for it not to be at least a bit confusing :-P. This PR can already be reviewed within the context of what it already promises (i.e. a new component that merges CreateObject and CreateCustom).

BHoM_UI/Components/oM/CreateObject2.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        /*************************************/
+        /**** Constructors                ****/
+        /*************************************/
+
+        public CreateObject2Caller() : base()
+        {
+            List<MemberInfo> possibleItems = new List<MemberInfo>();
+            possibleItems.AddRange(Engine.Reflection.Query.BHoMTypeList().Where(t => t?.GetInterface("IImmutable") == null));
+            possibleItems.AddRange(Engine.UI.Query.CreateItems());
```
epignatelli: All the methods that provide the possible items are in the [SearchItem](https://github.com/BHoM/BHoM_UI/blob/master/UI_Engine/Query/SearchItems.cs) method. Would it be convenient to move this logic there as well?

This would keep the list of possible items consistent between the global menu and the local menus.
https://github.com/BHoM/BHoM_UI/blob/master/BHoM_UI/Global/SearchMenu.cs#L84

UI_Engine/Compute/Organise.cs
```diff
@@ -100,6 +122,26 @@ public static partial class Compute
...
+        private static List<string> SplitPath(string path, List<string> toSkip)
+        {
+            if (path.EndsWith(")"))
+            {
+                return path.Split('.').Except(toSkip).ToList();
+            }
+            else
+            {
+                string[] split = path.Split('.');
+                return split.Concat(new string[] { split.Last() + "(...)" }).ToList();
```
epignatelli: Super minor - we might use curly braces to maintain the transparency of the object initialiser syntax?
```suggestion
                return split.Concat(new string[] { split.Last() + "() {...}" }).ToList();
```

BHoM_UI/Components/oM/CreateObject2.cs
adecler: Possibly. I would keep that for another PR though as none of the current components are currently going that route.

UI_Engine/Compute/Organise.cs
adecler: Not sure. I feels it would be more confusing to a non-developer.

adecler:
This is now ready for review and merging

BHoM_UI/Components/oM/CreateObject.cs
```diff
@@ -53,9 +57,157 @@ public class CreateObjectCaller : MethodCaller
...
+
+            return true;
+        }
+
+
+        /*************************************/
+        /**** Private Methods             ****/
+        /*************************************/
+
+        protected Func<object[], object> CreateConstructor(Type type, List<ParamInfo> parameters)
```
epignatelli: In the light of the migration of all the methods to construct delegates (https://github.com/BHoM/BHoM_Engine/blob/master/Reflection_Engine/Convert/ToFunc.cs) maybe `BH.Engine.Reflection.Create` is a better place for this?

epignatelli:
As a final comment, I think that we can address the comments in a separate pr, especially to unblock the adapter refactoring. Thus, I am approving, but not merging to have a coordination chat before that @adecler

al-fisher:
This is looking good - massively going to improve both user and coding experience I think.

One specific comment I have which I think _will_ be good to include in this PR if possible is the way in which the default constructor is represented in the menu.

Is currently showing up as `BH.oM.Structure.Elements.Bar(...)` as an example.
I think this is likely confusing for new users - as you do not necessarly know what the defining properties of an object are in advance.

Suggest we try and explicitly list out the defining properties as possible inputs. (Knowing that they can actually be removed or left as default).

`BH.oM.Structure.Elements.Bar(Node startNode, Node endNode, ISectionProperty sectionProperty, double orientationAngle, BarRelease release, BarFEAType fEAType,  Constraint4DOF support, Offset offset)`

For bar example will look not too dissimilar to current Engine.Create exposing and mapping directly across many of the object initialiser syntax. a method which we will remove. See screen shot below.

Also as these can get long - I do really like the ___and "x" more inputs___ added to the end where needed.


image

al-fisher:
@adecler @epignatelli @alelom - what do you think of the above?

alelom:
I generally agree we should find a better representation for the `(...)` constructor.

> Suggest we try and explicitly list out the defining properties as possible inputs. (Knowing that they can actually be removed or left as default).

Problem with this is if you end up having a Create constructor and the `(...)` constructor cut to the same string, I think.

al-fisher:
> Problem with this is if you end up having a Create constructor and the (...) constructor cut to the same string, I think.

Don't forget we will be removing any and all Create methods that are simply exposing defining properties only - so there will be, by definition, no conflicts or generating the same string.

al-fisher:
> I guess the problem is that the menu is created when Grasshopper loads. At that time we do not yet mine the properties of each object?

hmmm ... right. Thanks @epignatelli.
@adecler will this be too much of an overhead at load time? Keen to expose the public properties as parameters here if at all possible.

So there are alternate options above - but personally I do like the idea of not labouring the distinction between creating an object through selecting a Create method or reflected public properties through syntax. Is not important to user here - and may confuse. Interested in thoughts though?

adecler:
Eduardo is right that gathering the properties would be an overhead. But more importantly to me the distinction between the two type of constructors is actually quite important as they behave differently. One can have some of its input removed, the other cannot. Something you will only see when zooming on the component. Having a clear visual indicator of this difference in the menu is therefore important to avoid guessing work in my opinion. Alessio's point of the final method string looking the same due to the cutting of long parameter chain is also very relevant. 

@epignatelli , I like you suggestion on using the context menu to allow adding the base properties instead of providing them by default. Agreed that it would be better to do this on a separate PR though so we can merge this and unblock the PR from @alelom .

Let's catch up in your morning before merging this.

UI_Engine/Convert/ToBHoM.cs
```diff
@@ -35,15 +35,23 @@ public static partial class Convert
 
         [Input("property", "The system property to convert to bhom")]
         [Output("parameter", "The bhom parameter used in the bhom abstract syntax")]
-        public static oM.UI.ParamInfo ToBHoM(this PropertyInfo property)
+        public static oM.UI.ParamInfo ToBHoM(this PropertyInfo property, object instance = null)
```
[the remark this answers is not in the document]
adecler: It is for sure a rare case. I couldn't really find a better way of doing this (if we actually end up needing it as it is currently placeholder code for further development). I guess there could be an argument for having an additional 'config' parameter required in some conversion although I am not super keen on the idea. This one is purely for performance reason though. I am not a big fan of it so happy for any suggestion to go around it in a different way.

BHoM_UI/Components/oM/CreateObject.cs
adecler: Good point

BHoM_UI/Components/oM/CreateObject.cs
```diff
@@ -53,9 +57,157 @@ public class CreateObjectCaller : MethodCaller
...
+            }
+            else if (item is Type)
+            {
+                SelectedItem = item;
+                Type type = item as Type;
+                Name = type.Name;
+                Description = type.Description();
+
+                //object instance = Activator.CreateInstance(type);  // Potentially for later
+                string[] excluded = new string[] { "BHoM_Guid" };
```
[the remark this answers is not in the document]
adecler: I like that.

epignatelli:
I think that https://github.com/BHoM/BHoM_UI/pull/147#discussion_r341051725 is a bigger obstacle, because it can scary people - they don't know what to do when they are overwhelmed by all the properties.

For this other issue, I agree that the initialiser provides a different experience compared to the other methods. Despite that, the user does not yet know that when he/she's choosing the entry from the list for the first time.
I do not agree with conforming the initialiser string to the others in all its aspects, because the second time the user gets there, he/she cannot distinguish it from the rest.
I do agree that it should provide the same level of info though, like the available properties. But not at the cost of doubling the loading time.

al-fisher:
Agreed @epignatelli - good summary above. 

on this:
> I do agree that it should provide the same level of info though, like the available properties. But not at the cost of doubling the loading time.

Thinking your idea of curly braces could be a subtle but distinct way of clarifying

adecler:
To be honest I don't really see the point of listing all the properties explicitly while the idea is always the same: all properties are used as input.

al-fisher:
When you are creating an object through the UI you do not necessarily know what the defining properties are.
As a new user, or even as a returning user, I think is important to see side by side the different input options when you are creating your object. 
To use our favourite Bar as an example! 😄 
Seeing the options of BarbyStartAndEndNode and BarbyLine next to each other is important as can inform design of the workflow and any other objects you might need to create on the canvas.
The menus are actually a really helpful way of exploring what is possible.
Dos that make sense?

adecler:
It does. So you're saying this helps when people don't know what properties an object is made of and there are multiple ways of building an object. 

My concern was that it would not be as obvious that we are dealing with a variable input component anymore (hence the `...`) but this is a one time learning aspect versus a full time practical issue so I am with you now.

al-fisher:
I like your identification of “one time learning aspect versus a full time practical issue“.
Imagining there are a few ideas like that that we could highlight across the UI as we continue to refine. And therefore design best/most appropriate way to support both.

One idea I just had - is around icons. We obviously only want one component!! (that was this whole task!!) but the wee icons in the search list could subtly reference the difference between variable input component or not. Maybe? An option in addition to the written syntax anyway.

adecler:
Hmmm, the issue I have with that idea is that the icons in the search are an indicator of the component from the task bar used in each case. This might be confusing if the search shows an icon that is not actually in the task bar. I guess we could make a small change but 1) it might be missed by some people 2) others will consider as a different icon altogether.

al-fisher:
Yep fair point - if subtle, then may well defeat the point. Icon are small in the menu here too.

PR #164, opened 2019-11-13
Backward compatibility for removed Engine.Create methods

### LINKED ISSUE #163: Backward compatibility for removed Engine.Create methods
#### Description:
Most of the trivial Engine.Create methods will be removed in favour of the auto-generated constructors. The components needs to be able to update itself from a deleted without input from the user.

adecler:

Closes #163 

After an attempt of solving this problem within the Versioning_Toolkit, I ended up taking care of it directly in the BHoM_UI. The main reason for it is the lack of return type stored in the json version of a MethodInfo.  I tried to work with method Name and namespace to identify the return type but there was a few too many exceptions for my liking. Solving it here felt more robust and is after all a UI specific versioning issue. This doesn't change the fact that I will add a support for explicitly deprecated method in the versioning engine as planned. 

### Test files
- Create some CreateObject components (make sure they use a specific Engine.Create method) and save the file. 
- Comment out the corresponding Create methods in the Engine and recompile
- Reopened the saved script file and make sure the component is still working and now has the ability to remove inputs and an input  selector menu.

### Additional comments
For the input selection menu to remove pre-existing inputs properly, make sure to use the [corresponding PR in Grasshopper_Toolkit](https://github.com/BHoM/Grasshopper_Toolkit/pull/437). For Dynamo, the equivalent fix has been pushed to [the pending PR](https://github.com/BHoM/Dynamo_Toolkit/pull/179).

### DIFF
--- a/BHoM_UI/Components/oM/CreateObject.cs
+++ b/BHoM_UI/Components/oM/CreateObject.cs
@@ -32,6 +32,7 @@ using System.Linq.Expressions;
 using BH.oM.UI;
 using BH.Engine.UI;
 using System.Windows.Forms;
+using BH.oM.Base;
 
 namespace BH.UI.Components
 {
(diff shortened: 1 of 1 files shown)


al-fisher:
> Solving it here felt more robust and is after all a UI specific versioning issue. This doesn't change the fact that I will add a support for explicitly deprecated method in the versioning engine as planned.

Yep. Makes sense @adecler - thanks 

Also - I don't know if you have already explicitly added UI deprecation as an additional use case to versioning going forward (in addition to the namespace, type, method, internal property cases etc.)
Will be good to add and summarise - there will always continue to be tweaks to the UI going forward which we will need to support backwards compatibility for.  
However - agree there is significant difference between Versioning of oM and Engine, than rest of BHoM. With UI potentially requiring more responsibility for future changes in UI itself and Parent Adapter etc. Good to clarify all those potential cases and the lines between them.

Issue #166, opened 2019-11-14
Delete accepts only FilterRequest + to be renamed to Remove


alelom:
- Delete needs to reflect the change in the BHoM_Adapter refactoring: it needs to be renamed as `Remove`.

- It must accept any `IRequest` and not only `FilterRequest`

PR #167, opened 2019-11-14
Aligning with changes in BHoMAdapter Refactoring Level 04

alelom:
### NOTE: Depends on 

https://github.com/BHoM/BHoM_Engine/pull/1371   
https://github.com/BHoM/BHoM_Adapter/pull/164
   
Closes https://github.com/BHoM/BHoM_UI/issues/171
Closes https://github.com/BHoM/BHoM_UI/issues/166

### Changelog

- Added PushType/PullType to the respective method inputs
- Delete action renamed to Remove
- Adapter actions now accepts IRequest 
- General tidy up of UI methods for adapter actions

### DIFF
--- a/BHoM_UI/BHoM_UI.csproj
+++ b/BHoM_UI/BHoM_UI.csproj
@@ -67,6 +67,9 @@
       <HintPath>..\..\BHoM_Engine\Build\Library_Engine.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="MongoDB.Bson, Version=2.4.4.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\MongoDB.Bson.2.4.4\lib\net45\MongoDB.Bson.dll</HintPath>
+    </Reference>
     <Reference Include="PresentationCore">
       <Private>False</Private>
     </Reference>
@@ -116,6 +119,10 @@
       <Private>False</Private>
     </Reference>
     <Reference Include="System.Xml">
+      <Private>False</Private>
+    </Reference>
+    <Reference Include="Versioning_Engine">
+      <HintPath>..\..\BHoM_Engine\Build\Versioning_Engine.dll</HintPath>
       <Private>False</Private>
     </Reference>
     <Reference Include="WindowsBase">
@@ -224,6 +231,9 @@
       <Name>UI_oM</Name>
       <Private>False</Private>
     </ProjectReference>
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
--- a/BHoM_UI/BHoM_UI.csproj
+++ b/BHoM_UI/BHoM_UI.csproj
@@ -30,6 +30,10 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Adapter_oM">
+      <HintPath>..\..\BHoM_Adapter\Build\Adapter_oM.dll</HintPath>
+      <Private>False</Private>
+    </Reference>
     <Reference Include="BHoM">
       <HintPath>..\..\BHoM\Build\BHoM.dll</HintPath>
       <Private>False</Private>
--- a/BHoM_UI/BHoM_UI.csproj
+++ b/BHoM_UI/BHoM_UI.csproj
@@ -125,7 +125,7 @@
     <Compile Include="Components\Adapter\Move.cs" />
     <Compile Include="Components\Adapter\Execute.cs" />
     <Compile Include="Components\Adapter\Pull.cs" />
-    <Compile Include="Components\Adapter\Delete.cs" />
+    <Compile Include="Components\Adapter\Remove.cs" />
     <Compile Include="Components\Adapter\Push.cs" />
     <Compile Include="Components\Engine\Explode.cs" />
     <Compile Include="Components\Engine\GetInfo.cs" />
--- a/BHoM_UI/Components/Adapter/Move.cs
+++ b/BHoM_UI/Components/Adapter/Move.cs
@@ -69,9 +69,15 @@ namespace BH.UI.Components
         [Input("active", "Execute the Move")]
         [Output("success", "Define if the Move was successful")]
         public static bool Move(BHoMAdapter source, BHoMAdapter target, IRequest request = null, 
-            PullType pullType = PullType.AdapterDefault, Dictionary<string, object> pullConfig = null,
-            PushType pushType = PushType.AdapterDefault, Dictionary<string, object> pushConfig = null, bool active = false)
+            PullType pullType = PullType.AdapterDefault, ActionConfig pullConfig = null,
+            PushType pushType = PushType.AdapterDefault, ActionConfig pushConfig = null, bool active = false)
         {
+            if (source == null || target == null)
+            {
+                Engine.Reflection.Compute.RecordError("Adapter input cannot be null.");
+                return false;
+            }
+
             // ---------------------------------------------//
             // Mandatory Adapter Action set-up              //
             //----------------------------------------------//
--- a/BHoM_UI/Components/Adapter/Move.cs
+++ b/BHoM_UI/Components/Adapter/Move.cs
@@ -67,13 +67,24 @@ namespace BH.UI.Components
         [Input("pushConfig", "Config options for the Push to Target adapter")]
         [Input("active", "Execute the Move")]
         [Output("success", "Define if the Move was successful")]
-        public static bool Move(BHoMAdapter source, BHoMAdapter target, IRequest request = null, Dictionary<string, object> pullConfig = null, Dictionary<string, object> pushConfig = null, bool active = false)
+        public static bool Move(BHoMAdapter source, BHoMAdapter target, IRequest request = null, 
+            PullOption pullOption = PullOption.Unset, Dictionary<string, object> pullConfig = null, 
+            PushOption pushOption = PushOption.Unset, Dictionary<string, object> pushConfig = null, bool active = false)
         {
+            // ---------------------------------------------//
+            // Mandatory Adapter Action set-up              //
+            //----------------------------------------------//
+            // The following are mandatory set-ups to be ALWAYS performed 
+            // before the Adapter Action is called,
+            // whether the Action is overrided at the Toolkit level or not.
+
             if (request == null)
                 request = new FilterRequest();
 
+            //----------------------------------------------//
+
             if (active)
-                return source.Move(target, request, pullConfig, pushConfig);
+                return source.Move(target, request, pullOption, pullConfig, pushOption, pushConfig);
             else
                 return false;
         }
--- a/BHoM_UI/Components/Adapter/Move.cs
+++ b/BHoM_UI/Components/Adapter/Move.cs
@@ -85,6 +85,10 @@ namespace BH.UI.Components
             // before the Adapter Action is called,
             // whether the Action is overrided at the Toolkit level or not.
 
+            // If unset, set the actionConfig to a new ActionConfig.
+            pullConfig = pullConfig == null ? new ActionConfig() : pullConfig;
+            pushConfig = pushConfig == null ? new ActionConfig() : pushConfig;
+
             if (request == null)
                 request = new FilterRequest();
 
--- a/BHoM_UI/Components/Adapter/Move.cs
+++ b/BHoM_UI/Components/Adapter/Move.cs
@@ -60,12 +60,14 @@ namespace BH.UI.Components
         /**** Public Method               ****/
         /*************************************/
 
-        [Description("Copy objects from a Source adapter to a Target adapter")]
+        [Description("Adapter Action 'Move': copies objects from an external software to another. \nThe objects do not pass through the UI to avoid performance bottleneck.")]
         [Input("source", "Adapter the data is copied from")]
         [Input("target", "Adapter the data is copied to")]
         [Input("request", "Filter on the objects to pull (default: get all)")]
-        [Input("pullConfig", "Config options for the Pull from Source adapter")]
-        [Input("pushConfig", "Config options for the Push to Target adapter")]
+        [Input("pullConfig", "Configuration for the Pull actioned by the Move. You can input an ActionConfig (it contains the configs common to all Toolkits); \n" +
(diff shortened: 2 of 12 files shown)


adecler:
Just noting that the `BHoM_UI.ChekCore` test fails because Dynamo doesn't yet have a corresponding PR.

epignatelli:
Is this ready to review, or should I wait?

BHoM_UI/Components/Adapter/Move.cs
```diff
@@ -67,13 +68,30 @@ public class MoveCaller : MethodCaller
         [Input("pushConfig", "Config options for the Push to Target adapter")]
         [Input("active", "Execute the Move")]
         [Output("success", "Define if the Move was successful")]
-        public static bool Move(BHoMAdapter source, BHoMAdapter target, IRequest request = null, Dictionary<string, object> pullConfig = null, Dictionary<string, object> pushConfig = null, bool active = false)
+        public static bool Move(BHoMAdapter source, BHoMAdapter target, IRequest request = null, 
+            PullType pullType = PullType.AdapterDefault, ActionConfig pullConfig = null,
```
epignatelli: Is there a reason we are defaulting these variables here and not in the BHoMAdapter itself?

BHoM_UI/Components/Adapter/Pull.cs
```diff
@@ -82,17 +83,45 @@ public override object Run(object[] inputs)
...
-        [Input("request", "Filter on the objects to pull")]
+        [Input("request", "Connect a type of object to Pull objects of that type." +
+            "Connect a Request to pull only objects that satisfy a certain rule.")]
         [Input("config", "Pull config")]
         [Input("active", "Execute the pull")]
-        [Output("objects","Objects pulled")]
-        public static IEnumerable<object> Pull(BHoMAdapter adapter, IRequest request = null, Dictionary<string, object> config = null, bool active = false)
+        [Output("objects", "Objects pulled")]
+        public static IEnumerable<object> Pull(BHoMAdapter adapter, IRequest request = null, 
+            PullType pullType = PullType.AdapterDefault, ActionConfig actionConfig = null,
```
epignatelli: As above

BHoM_UI/Components/Adapter/Pull.cs
```diff
@@ -82,17 +83,45 @@ public override object Run(object[] inputs)
...
         [Input("config", "Pull config")]
         [Input("active", "Execute the pull")]
-        [Output("objects","Objects pulled")]
-        public static IEnumerable<object> Pull(BHoMAdapter adapter, IRequest request = null, Dictionary<string, object> config = null, bool active = false)
+        [Output("objects", "Objects pulled")]
+        public static IEnumerable<object> Pull(BHoMAdapter adapter, IRequest request = null, 
+            PullType pullType = PullType.AdapterDefault, ActionConfig actionConfig = null, 
+            bool active = false)
         {
+            IRequest actualRequest = null;
```
epignatelli: As above

BHoM_UI/Components/Adapter/Pull.cs
```diff
@@ -82,17 +83,45 @@ public override object Run(object[] inputs)
...
+            // ---------------------------------------------//
+            // Mandatory Adapter Action set-up              //
+            //----------------------------------------------//
+            // The following are mandatory set-ups to be ALWAYS performed 
+            // before the Adapter Action is called,
+            // whether the Action is overrided at the Toolkit level or not.
+ 
             if (request == null)
-                request = new FilterRequest();
+                actualRequest = new FilterRequest();
```
epignatelli: Same here. Wouldn't this and all the below, apart from the null check on the `adapter`, be safer be in the BHoMAdapter?

BHoM_UI/Components/Adapter/Push.cs
```diff
@@ -63,15 +64,27 @@ public class PushCaller : MethodCaller
...
-        [Input("config", "Push config")]
+        [Input("pushType", "Push type. Connect the enum PushType for all the alternatives.")]
+        [Input("config", "Push config. Dictionary<string,object> containing any additional data to be used for the Push.")]
         [Input("active", "Execute the push")]
-        [MultiOutput(0, "objects", "Objects that have been pushed (with potentially additional information stored in their CustomData to reflect the push)")]
+        [MultiOutput(0, "objects", "Objects that have been pushed.\nThese objects may be different from the input objects (e.g. their correspondent external software id may be stored in their CustomData).")]
         [MultiOutput(1, "success", "Define if the push was sucessful")]
-        public static Output<List<IObject>, bool> Push(BHoMAdapter adapter, IEnumerable<IObject> objects, string tag = "", Dictionary<string, object> config = null, bool active = false)
+        public static Output<List<object>, bool> Push(BHoMAdapter adapter, IEnumerable<object> objects, string tag = "",
+            PushType pushType = PushType.AdapterDefault, ActionConfig actionConfig = null,
```
epignatelli: Ditto

BHoM_UI/Components/Adapter/Remove.cs
```diff
@@ -46,32 +47,45 @@ public class DeleteCaller : MethodCaller
...
+            }
+
+            // ---------------------------------------------//
+            // Mandatory Adapter Action set-up              //
+            //----------------------------------------------//
+            // The following are mandatory set-ups to be ALWAYS performed 
+            // before the Adapter Action is called,
+            // whether the Action is overrided at the Toolkit level or not.
+
+            if (request == null)
```
epignatelli: Ditto

BHoM_UI/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="MongoDB.Bson" version="2.4.4" targetFramework="net452" />
```
epignatelli: Why is this necessary? Referring the `Serialization_Engine.dll` is not enough?

BHoM_UI/BHoM_UI.csproj
```diff
@@ -63,6 +67,9 @@
       <HintPath>..\..\BHoM_Engine\Build\Library_Engine.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="MongoDB.Bson, Version=2.4.4.0, Culture=neutral, processorArchitecture=MSIL">
```
epignatelli: Why is this necessary? Referring the `Serialization_Engine.dll` is not enough?

BHoM_UI/Templates/Caller.cs
```diff
@@ -241,22 +244,47 @@ public virtual bool Read(string json)
                 // New serialisation, we stored a CustomObject with SelectedItem, InputParams and OutputParams
                 object backendElement;
                 if (component.CustomData.TryGetValue("SelectedItem", out backendElement))
+                {
+                    if (backendElement == null)
+                    {
+                        BsonDocument bson = Engine.Serialiser.Convert.ToBson(json);
```
epignatelli: I don't think this should be handled here. We have the `Versioning_Engine`, right?

BHoM_UI/BHoM_UI.csproj
alelom: I think that's been added by @adecler's changes.

BHoM_UI/Templates/Caller.cs
alelom: I think that's been added by @adecler's changes.

BHoM_UI/packages.config
alelom: I think that's been added by @adecler's changes.

BHoM_UI/Components/Adapter/Pull.cs
alelom: As visible, I didn't add this in for this refactoring. The reason it's there is to assure we have always have a value passed to all Adapter methods.
This cannot be assured in the Adapter because we allow the override of the Adapter Actions.

BHoM_UI/Components/Adapter/Move.cs
```diff
@@ -67,13 +68,30 @@ public class MoveCaller : MethodCaller
...
+                return false;
+            }
+
+            // ---------------------------------------------//
+            // Mandatory Adapter Action set-up              //
+            //----------------------------------------------//
+            // The following are mandatory set-ups to be ALWAYS performed 
+            // before the Adapter Action is called,
+            // whether the Action is overrided at the Toolkit level or not.
+
```
alelom: This is the reason why we want to have some initialisation in the UI itself rather than later in the adapter.
We allow for Action overloading and we want to assure that certain initialisation are done no matter the overload.

BHoM_UI/Components/Adapter/Move.cs
alelom: As per our chat and as further explained [here](https://github.com/BHoM/BHoM_UI/pull/167/files/40e23834c3eb21462d165709731bbffa938f3866#r354452867)
there are certain initialisation that must be assured before the Adapter methods (aka Actions) are called, because we allow for them to be overloaded.

BHoM_UI/Components/Adapter/Pull.cs
alelom: See my comment here https://github.com/BHoM/BHoM_UI/pull/167#discussion_r354453473

BHoM_UI/Components/Adapter/Pull.cs
alelom: See my comment here #167 (comment)

BHoM_UI/Components/Adapter/Pull.cs
alelom: See my comment here #167 (comment)

BHoM_UI/Components/Adapter/Push.cs
alelom: See my comment here #167 (comment)

BHoM_UI/Components/Adapter/Remove.cs
alelom: See my comment here #167 (comment)

adecler:
@epignatelli , some of the versioning needs to happen on the level of the UI, hence the link to Mongo.Bson.

alelom:
> `+ $exception {"Unable to cast object of type 'MongoDB.Bson.BsonNull' to type 'MongoDB.Bson.BsonDocument'."} System.InvalidCastException`
> 
> This PR is messing with the CustomObjects,
> Try creating a customObject with properties, then copy and paste it or close and re-open grasshopper.
> Afterwards the assigned properties disappear.
> image

Thanks for this @kThorsager. 
I had a look, I think this could be linked to one of the changes done by @adecler, either on:
- [`Caller.cs`](https://github.com/BHoM/BHoM_UI/pull/167/files#diff-57fe9b87f51a6223525a669c505b65e6)
- [`CreateObject.cs`](https://github.com/BHoM/BHoM_UI/pull/167/files#diff-59b7ada80c65d7f337cc8717427ab4aa)
@adecler could you please help me bottoming this out?

adecler:
Thanks @kThorsager , good (and critical!) catch 😅 .

I have fixed it in an additional commit. This issue was that I had forgotten to handle the case where `SelectedItem` is null before serialising it (which is always the case for `CreateCustom`). 

Should all be fine now but let me know if you still find problems.

alelom:
Thanks for this @adecler!

BHoM_UI/Components/Adapter/Pull.cs
```diff
@@ -80,19 +81,43 @@ public override object Run(object[] inputs)
...
         {
+            // ---------------------------------------------//
+            // Mandatory Adapter Action set-up              //
+            //----------------------------------------------//
+            // The following are mandatory set-ups to be ALWAYS performed 
+            // before the Adapter Action is called,
+            // whether the Action is overrided at the Toolkit level or not.
+
+            // If unset, set the actionConfig to a new ActionConfig.
+            actionConfig = actionConfig == null ? new ActionConfig() : actionConfig;
```
epignatelli: All of this happens because we allow users to override the `Pull` method, in this case. If we prevent that, we can avoid spreading a lot of checks in other projects. This is explained in the Adapter pr https://github.com/BHoM/BHoM_Adapter/pull/164

BHoM_UI/Components/Adapter/Pull.cs
```diff
@@ -80,19 +81,43 @@ public override object Run(object[] inputs)
...
+            // Always assure there is a Request. Allow to input a Type to generate a FilterRequest.
+            IRequest actualRequest = null;
+
             if (request == null)
-                request = new FilterRequest();
+                actualRequest = new FilterRequest();
+            else if (request is Type)
+                actualRequest = BH.Engine.Data.Create.FilterRequest((Type)request, "");
+            else if (typeof(IRequest).IsAssignableFrom(request.GetType()))
+                actualRequest = request as IRequest;
```
epignatelli: Just to reinforce, this code is here because we allow the override of the `Pull` method and. because of that, we are not sure the `base.Pull` will be called - thus, we cannot enforce this logic.

The more I read the code, the more I think that the `Pull` method should not be `virtual`, and anything the user wants to do, can be done in a `BeforePull` and `AfterPull` methods.

BHoM_UI/Components/Adapter/Pull.cs
adecler: We need to be able to override the `Pull` in some of the lower level adapters

alelom: I agree with both - a mechanism like the one suggested by @epignatelli in https://github.com/BHoM/BHoM_Adapter/pull/164 would still be advisable in my mind, so we can limit overrides to only the _strictly neededed_ cases.

adecler: > so we can limit overrides to only the strictly neededed cases.

Please no! There is nothing more painful when you are hacking with someone else code than to realise that the stuff you want to change is not overridable.

alelom: @adecler that's why I said "limit" and didn't say "disallow completely"!
Could you please confirm you would oppose a strategy where we have:
- overridable "Before[SomeActionName]" and "After[SomeActionName]" (where ActionName is Push/Pull/etc), that run before and after each individual adapter action, and
- overridable Actions (in which the override is permitted, but discouraged unless strictly needed)?

adecler: I would. Anything that doesn't allow to completely override the adapter actions is a step back for me.  `Before` and `After` overridable methods still means that you have no say on code being executed in the middle. There are better ways to guide new contributors without limiting advanced developers than restricting access to code.

PR #173, opened 2019-12-11
Provide object descriptions if they're available on default create components

### LINKED ISSUE #172: BHoM_UI: Allow object property descriptions to be used as input descriptions
Following the implementation of default create components, they do not contain any description of what the input is or documentation to help users work with those components.

This was previously provided by the `InputAttribute` for code-made create components.

However, now we are pushing for all BHoM Object properties to include `Description` attributes to describe them better to users. This description attribute, if available, should be used in the default create components as the documentation in the same way `InputAttribute` descriptions are used on coder-created methods.

Putting on 3.0 Release milestone as this is primarily a documentation bug that I think should be solved before 3.0 release.

FraserGreenroyd:
### Depends on
https://github.com/BHoM/BHoM_Engine/pull/1388

Fixes #172 

### Test files
Load any create component into a UI using the default create component rather than a coder-created one from an engine

### Additional comments
@adecler it was found during the investigation of this issue that you added `.PropertyType` to the `property` variable in the file which is changed, so I'm not 100% that this is the right fix. It fixes the problem I am trying to solve, but am cautious that it might thus mess something else up. As such, I would like to wait for a review from you before merging to double check just in case :smile: I'm asking for a review from the other UI developers to just make sure they're happy with Excel/GH/Dynamo points of view but final approval should come from @adecler

### DIFF
--- a/UI_Engine/Convert/ToBHoM.cs
+++ b/UI_Engine/Convert/ToBHoM.cs
@@ -41,7 +41,7 @@ namespace BH.Engine.UI
             {
                 Name = property.Name,
                 DataType = property.PropertyType,
-                Description = property.PropertyType.IDescription(),
+                Description = property.IDescription(),
                 Kind = ParamKind.Input
             };
 


awakeman:
Note: Here's an example showing the difference this PR makes for `BH.oM.Gradient` that has the following code:

```csharp
public class Gradient : BHoMObject
{
    /***************************************************/
    /****            Public Properties              ****/
    /***************************************************/

    [Description("A SortedDictionary of 'Color's using a 'decimal' between 0 and 1 as Key. Is used to get a interpolated 'Color' from a value between 0 and 1")]
    public SortedDictionary<decimal, Color> Markers { get; set; } = new SortedDictionary<decimal, Color>();

    /***************************************************/

}
```


Before:
image

After:
image

IsakNaslundBh:
> This creates no issues in Grasshopper.
> It feels sensible, but as @FraserGreenroyd said, I ma not sure why `PropertyType` was there in the first place.

I guess it was to get any potential description of the class itself, rather than the property, which also would make sense. Think the change in here should be done though, as well as the following in the reflection engine to make sure we pick up both class descriptions as well as property descriptions:

image

adecler:
I think the change here is fine **_as long as_** this PR is paired with a change in the `Description` method itself. Right now, we are loosing the important information of the type of object you need to provide as an input. See the typical example of bars and nodes once we use this PR:

image

This is about as useless as it can be 😋 . Someone new to the BHoM would have no idea what to create next or search for in order to provide that `StartNode`. 

I sort of agree with @IsakNaslundBh about his deletion of the trivial description line but his change doesn't make sense to me (might be my holiday brain) as it would provide the description of the containing class instead of the property itself. So in the example above, you would have the description of a `Bar` when hovering over `StartNode` 😕 .

I'll do a quick PR on the Relection_Engine. I'll mark this as `do-not-merge` in the mean time.

FraserGreenroyd:
@adecler can I confirm if you're happy with this PR being merged when suitably tested by @epignatelli @awakeman @IsakNaslundBh et. al., along with the PR on [Reflection_Engine](https://github.com/BHoM/BHoM_Engine/pull/1388), or would you want additional testing/review? :smile:

IsakNaslundBh:
> I sort of agree with @IsakNaslundBh about his deletion of the trivial description line but his change doesn't make sense to me (might be my holiday brain) as it would provide the description of the containing class instead of the property itself.

As discussed over skype, was me not looking it through it enough. What you did on the PR branch was what I intended!

PR #174, opened 2019-12-12
Make sure that `ComponentRequest` has a location when menu is navigated via keyboard inputs

epignatelli:

This pr support the fixes implement in https://github.com/BHoM/Grasshopper_Toolkit/pull/445 bt does not depend on it.

### Test files

You can test this pr in tandem with the associated Grasshopper_Toolkit one

### Changelog

- `NotifySelection` now always takes a locations
- Make sure that when using the keyboard to navigate the BHoM menu, the component is not created at the mouse cursor location, but at the location of the top left corner of the menu


epignatelli:
Love the CheckCore ❤️ @FraserGreenroyd @al-fisher

PR #180, opened 2020-01-08
Improved error message when exception reached Caller.Run

### LINKED ISSUE #179: BHoM_UI: Remove boilerplate text for errors?
Currently when an error is raised somewhere in the process and returned to the user, the UI adds the text "1. This component failed to run properly. Are you sure you have the correct type of inputs?
Check their description for more details. Here is the error provided by the method:" before the actual message is received. I personally feel this is more in the way for reading the actual error message and not that helpful. Also not sure if it helps new-starters as it generally will not be able to give any guidance of what is wrong. I would vote for removing this text and only returning the error message from the method. Would be glad for inputs from others here as well @FraserGreenroyd @adecler @al-fisher @alelom @epignatelli @JosefTaylor @enarhi @peterjamesnugent 

image

adecler:

Closes #179
Closes https://github.com/BHoM/BHoM_Engine/issues/1375

### Additional comments
Since the code was already on my machine, I thought I might as well pull request it. Don't take this as  the final say on how to solve #179 though. I am happy to iterate on it as needed.

### DIFF
--- a/BHoM_UI/Templates/Caller.cs
+++ b/BHoM_UI/Templates/Caller.cs
@@ -104,8 +104,26 @@ namespace BH.UI.Templates
             }
             catch (Exception e)
             {
-                RecordError(e, "This component failed to run properly. Are you sure you have the correct type of inputs?\n" +
-                                 "Check their description for more details. Here is the error provided by the method:\n");
+                string message = "This component failed to run properly. Are you sure you have the correct type of inputs?\n" +
+                                 "Check their description for more details. Here is the error provided by the method:\n   ";
+
+                if (e.InnerException != null)
+                    message += e.InnerException.Message;
+                else
+                    message += e.Message;
+
+                List<string> stack = e.StackTrace.Split(new char[] { '\n' })
+                    .Where(x => x.Contains(" BH."))
+                    .Take(2)
+                    .ToList();
+
+                if (stack.Count > 0)
+                    message += "\nHappened " + stack[0].Trim();
+                if (stack.Count > 1)
+                    message += "\n   called " + stack[1].Trim();
+
+                Engine.Reflection.Compute.RecordError(message);
+
                 return false;
             }
 
--- a/BHoM_UI/Templates/Caller.cs
+++ b/BHoM_UI/Templates/Caller.cs
@@ -104,8 +104,7 @@ namespace BH.UI.Templates
             }
             catch (Exception e)
             {
-                string message = "This component failed to run properly. Are you sure you have the correct type of inputs?\n" +
-                                 "Check their description for more details. Here is the error provided by the method:\n   ";
+                string message = "This component failed to run properly.\n- Error: "; 
 
                 if (e.InnerException != null)
                     message += e.InnerException.Message;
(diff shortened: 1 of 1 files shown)


FraserGreenroyd:
I suppose my question would be could we ever have a situation where the stack is deeper than 2? If not then I'm happy with this proposed solution to the problem presented, I think it will help with debugging a lot and help users tell us where problems are found massively!

alelom:
I like @adecler proposal!

> I suppose my question would be could we ever have a situation where the stack is deeper than 2? 

More than possible – my proposal with the 2 levels of stack was because I noticed it's usually enough to have a better understanding of the problem. Asking it to cover all the stack levels would make it essentially the same of a classic stack trace, and would be incompatible with our requirement of presenting not-too-scary, easy-to-understand errors.

FraserGreenroyd:
> > I suppose my question would be could we ever have a situation where the stack is deeper than 2?
> 
> More than possible – my proposal with the 2 levels of stack was because I noticed it's usually enough to have a better understanding of the problem. Asking it to cover all the stack levels would make it essentially the same of a classic stack trace, and would be incompatible with our requirement of presenting not-too-scary, easy-to-understand errors.

That seems fair, in which case I'm happy to approve this PR from my side :smile:

adecler:
> > I suppose my question would be could we ever have a situation where the stack is deeper than 2? 
> 
> More than possible – my proposal with the 2 levels of stack was because I noticed it's usually enough to have a better understanding of the problem. Asking it to cover all the stack levels would make it essentially the same of a classic stack trace, and would be incompatible with our requirement of presenting not-too-scary, easy-to-understand errors. 

This exactly. The stack can be decently long and cryptic. That's why I kept Alessio's idea to stick to 2 levels no matter what. The only difference is that I keep the first two BH levels instead of first two levels and then check that they are BH.

IsakNaslundBh:
Think this is a nice addition! I still would remove/move/shorten the starting text though, as currently the exception message (which still feels like the most important part, at least to me) is sort of hidden in between the default text and the stack trace.

I like the suggestion by @al-fisher here https://github.com/BHoM/BHoM_UI/issues/179#issuecomment-571619560 in moving the default text to after everything else. Then you get the useful information first followed by some simple suggestion on what could be wrong.

al-fisher:
See my proposal in this https://github.com/BHoM/BHoM_UI/issues/179#issuecomment-571947433 too giving info in order of specificity (most specific first)

adecler:
Based on @al-fisher comment, here's the new version:

image

rwemay:
This looks good to me - really helpful! Maybe going to far (and expanding the output text again!) but if it's possible to report the position in a list/loop it would be really helpful. Frequently it'll be one object in a list that's causing things to fail. Not sure if I'm explaining this very well, hope makes sense. Also feel free to ignore as it's beyond scope of this issue!

epignatelli:
@IsakNaslundBh @al-fisher would you mind reviewing according to the latest changes?

> This looks good to me - really helpful! Maybe going to far (and expanding the output text again!) but if it's possible to report the position in a list/loop it would be really helpful. Frequently it'll be one object in a list that's causing things to fail. Not sure if I'm explaining this very well, hope makes sense. Also feel free to ignore as it's beyond scope of this issue!

I think this has to be solved in each UI implementation, such as the Grasshopper_Toolkit, since it is the one keeping track of the iteration. in case it happens in an actual `for` loop, I think it's a matter of recording the error with that information.

PR #184, opened 2020-01-09
Fix slow GetProperty component

### LINKED ISSUE #182: BHoM_UI: Explode() is faster that GetProperty
As in the header, currently it is faster to retrieve a property of an object with the Explode component than with GetProperty - not sure if that is the desired behaviour (intuitively one would say GetProperty should be faster).

adecler:

Closes #182

### Test files

This is the computational time before fix:

image

And after the fix:

image

### DIFF
--- a/BHoM_UI/Components/Engine/GetProperty.cs
+++ b/BHoM_UI/Components/Engine/GetProperty.cs
@@ -87,8 +87,13 @@ namespace BH.UI.Components
             if (OutputParams.Count < 1)
                 return true;
 
-            OutputParams[0].DataType = result.GetType();
-            CompileOutputSetters();
+            Type type = result.GetType();
+            if (OutputParams[0].DataType != type)
+            {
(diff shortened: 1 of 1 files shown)


BHoM_UI/Components/Engine/GetProperty.cs
```diff
@@ -87,8 +87,13 @@ public bool SetOutputTypes(object result)
             if (OutputParams.Count < 1)
                 return true;
 
-            OutputParams[0].DataType = result.GetType();
-            CompileOutputSetters();
+            Type type = result.GetType();
+            if (OutputParams[0].DataType != type)
+            {
+                OutputParams[0].DataType = result.GetType();
```
pawelbaran: Could use `type` instead of `result.GetType()`

adecler: Ha, that was the initial plan. Seems like I got distracted at the wrong time there 😋

PR #186, opened 2020-01-10
Allow to remove outputs for the Explode component and show BHoM menu on wire drag-and-released event

### LINKED ISSUE #183: Allow to remove outputs for the Explode component
This should make the Explode component more practical to use when the user is only interested in a few of the object properties by keeping the size of the component to a minimum

### LINKED ISSUE #196: Show BHoM menu on wire drag-and-released event
When dragging a wire from a BHoM parameter (as an output), show a menu of available methods that accept that object as a input. Add a search box on the top of it.

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_Engine/pull/1437
This is not a strict dependency in term of code compilation but the change in the UI have highlighted a bug in the engine so better to test with that PR.

Closes #183 
Closes #196 

### Changelog
- Allow to remove outputs for the Explode component
- Show BHoM menu on wire drag-and-released event

### Additional comments

- Note that if you are doing a test in Excel, you will need this PR for the global menu to still work properly: https://github.com/BHoM/Excel_Toolkit/pull/190
- I am fixing two issues in this PR because those were the issues that had code change overlapping with the adapter level 4 PRs. So it was easier for me to store them and submit them as a batch.

### DIFF
--- a/UI_Engine/Query/Weight.cs
+++ b/UI_Engine/Query/Weight.cs
@@ -94,6 +94,101 @@ namespace BH.Engine.UI
         }
 
         /*************************************/
+
+        public static double Weight(this SearchItem item, Type constraint, bool isReturnType = true)
+        {
+            constraint = constraint.UnderlyingType().Type;
+
+            if (isReturnType)
+            {
+                object target = item.Item;
+                Type returnType = typeof(object);
+
+                if (target is MethodInfo)
+                    returnType = ((MethodInfo)target).ReturnType;
+                else if (target is ConstructorInfo)
+                    returnType = ((ConstructorInfo)target).DeclaringType;
+                else if (target is Type)
+                {
+                    if (item.CallerType.Name == "CreateTypeCaller")
+                        returnType = typeof(Type);
+                    else
+                        returnType = target as Type;
+                }
+
+                double callerWeight = item.Text.StartsWith("BH.oM") ? 1.0 : 0.1;
+                return callerWeight * DistanceWeight(returnType.UnderlyingType().Type, constraint);
+            }
+            else
+            {
+                object target = item.Item;
+
+                if (target is MethodBase)
+                {
+                    ParameterInfo[] parameters = ((MethodBase)target).GetParameters();
+                    if (parameters.Length == 0)
+                        return 0;
+                    else
+                        return parameters.Max(x => DistanceWeight(constraint, x.ParameterType.UnderlyingType().Type));
+                }
+                else if (target is Type)
+                {
+                    if (item.CallerType.Name == "CreateTypeCaller")
+                        return constraint == typeof(Type) ? 1 : 0;
+                    else
+                    {
+                        PropertyInfo[] properties = ((Type)target).GetProperties();
+                        if (properties.Length == 0)
+                            return 0;
+                        else
+                            return properties.Max(x => DistanceWeight(constraint, x.PropertyType.UnderlyingType().Type));
+                    }    
+                }
+                else
+                    return 0;
+            }
+        }
+
+
+        /*************************************/
+        /**** Private Methods             ****/
+        /*************************************/
+
+        private static double DistanceWeight(Type source, Type target)
+        {
+            if (source == target)
+                return 1;
+
+            if (target == typeof(object))
+                return 0.01;
+
+            if (!target.IsAssignableFrom(source))
+                return 0;
+
+            Type parent = source.BaseType;
+            IEnumerable<Type> toProcess = source.GetInterfaces();
+
+            for (int depth = 1; depth < 10; depth++)
+            {
+                List<Type> direct = toProcess.Except(toProcess.SelectMany(x => x.GetInterfaces()).Distinct()).ToList();
+                if (parent != null)
+                {
+                    direct = direct.Except(parent.GetInterfaces()).ToList();
+                    direct.Add(parent);
+                    parent = parent.BaseType;
+                }
+                if (direct.Contains(target))
+                    return 0.5 / depth;
+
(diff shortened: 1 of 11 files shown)


adecler:
See https://github.com/BHoM/Grasshopper_Toolkit/pull/453#issuecomment-577038038

adecler:
Now ready for review and merging

UI_Engine/Query/Weight.cs
```diff
@@ -94,5 +94,100 @@ public static double Weight(this SearchItem item, string[] search)
...
+                    return 0.5 / depth;
+
+                toProcess = toProcess.Except(direct);
+                if (toProcess.Count() == 0)
+                    break;
+            }
+
+            return 0;
+        }
+
```
epignatelli:
```suggestion

        /*************************************/

        private static Type TryGetPropertyType(this PropertyInfo property)
        {
            try
            {
                return property.PropertyType;
            }
            catch
            {
                return typeof(object);
            }
        }

```

UI_Engine/Query/Weight.cs
```diff
@@ -94,5 +94,100 @@ public static double Weight(this SearchItem item, string[] search)
...
+                {
+                    if (item.CallerType.Name == "CreateTypeCaller")
+                        return constraint == typeof(Type) ? 1 : 0;
+                    else
+                    {
+                        PropertyInfo[] properties = ((Type)target).GetProperties();
+                        if (properties.Length == 0)
+                            return 0;
+                        else
+                            return properties.Max(x => DistanceWeight(constraint, x.PropertyType.UnderlyingType().Type));
```
epignatelli:
```suggestion
                            return properties.Max(x => DistanceWeight(constraint, x.TryGetPropertyType().UnderlyingType().Type));
```

adecler:
@epignatelli , regarding #183, the idea is that if a user has gone through the process of manually selecting outputs, he will not want the explode component to update automatically. See this case for example:

Explode_SelectedOutputs

It is clear to me that the user will want to keep the output selection. AS always, I prefer to automate when there is 0% chance that this will make the wrong choice for the user. Any other case, I'd rather let the user choose.

In you example, you had the explode component turning red. This was due to the component trying to assign null to a `double` type (which is not nullable) . Not completely related to this PR but I pushed the fix here anyway to avoid the confusion.

PS: the structure engineer among you will notice that I left `Support` as a common property between bar and node. which is incorrect since one is is a 4DOF constraint while the other one is 6DOF. If the Node had a non-null support, we would have had the following error:

image

Which is expected behaviour in my book.

adecler:
@epignatelli , regarding point 1 of #196, I believe you are going to have that problem in quite a few places so can you test it further and raise a separate issue if you manage to cause other crashes. 

In the meantime, I have committed your modification to cover the urgent case

epignatelli:
> @epignatelli , regarding point 1 of #196, I believe you are going to have that problem in quite a few places so can you test it further and raise a separate issue if you manage to cause other crashes.

Thanks for the fix, I agree, there might be a lot of places when that will happen. I'll raise an issue as soon as I find one.

> Regarding point 2, I could go behind it on a personal user level. That being said, this is another feature that would be hidden behind a keyboard shortcut. I have already noticed that a lot of features were unknown by users in my office simply because it was hidden behind a menu, a keyboard shortcut or the need to zoom. So I am not too sure I want to add to the list. To help the decision, I looked into how it is done in Unreal Blueprint and they do it the same way as currently done in this PR (i.e. no need to press a key).

Yeah, I get the point, but I am not sure the solution is to expose it so strongly.  Maybe we can to have a media (a set of gifs, a BHoM primer, or a text) where we show the UI features?

@IsakNaslundBh @al-fisher @rwemay @alelom @FraserGreenroyd @awakeman maybe I am wrong, do you have an opinion on this last point?

al-fisher:
Ah - this merge conflict is caused by a file being actually deleted/renamed! Was wondering why! 
Sorry for that @adecler @epignatelli 
@FraserGreenroyd

al-fisher:
A wee one for us to fix tomorrow

FraserGreenroyd:
> A wee one for us to fix tomorrow

Unless @adecler fixes it overnight I'll get on this as a first action before sprint closure :smile:

adecler:
> Yeah, I get the point, but I am not sure the solution is to expose it so strongly. If the problem is the communication of features, maybe we should solve that one. Perhaps a media (a set of gifs, a BHoM primer, or a text) where we show the UI features?

So far, I only had a chance to talk to @IsakNaslundBh but he agreed that it is better not to have something like that hidden behind a keyboard shortcut. 

> Approving based on the fact that there is are no other opinions and the rest is fine.

In general, PRs should only be blocked with a change request due to bugs or not fully resolving the issues, not for aspects purely subject to opinion. Those should be raised as separate issues with the `question` label.

FraserGreenroyd:
The conflict should now be resolved, @epignatelli could you take another quick look and reapprove if you're happy - no code has been changed from what @adecler did so I do not anticipate any problems arising from this but you saw the functionality before this conflict so you're well placed to determine the resolution has not broken anything.

al-fisher:
Thanks @FraserGreenroyd @epignatelli

epignatelli:
> In general, PRs should only be blocked with a change request due to bugs or not fully resolving the issues, not for aspects purely subject to opinion. Those should be raised as separate issues with the `question` label.

This is a feature, it is clearly subject to opinion. Convincing the team is beneficial from my point of view. 
If the problem is that this pr is blocing further development, we should revise that instead.

Conflict resolved. I am merging.

PR #211, opened 2020-02-24
Adding ToNewVersion call for old component deserialisation

### LINKED ISSUE #210: BHoM_UI: Fails too much on old components
#### Description:

BHoM_UI does not send faulty methods to the Versioning_Engine if there are to grievous errors.
i.e. I open a script with old components in it, which are so serious that it displays errors before opening. This however also seems to cause the UI to be unable to pick up the errors it can handle, which causes the not to be sent to the Versioning_Engine. Hence all the components which should have been upgraded are set as magic Components.

Re-instancing the method is also unsuccessful as it is now saved as a magic component

This is set as critical as it is holding up the migration of `IElement`s and the name change of `IFragment`, which both would benefit greatly from Versioning.
#### Steps to reproduce
switch to this branch `Dimensional_oM698-Migrate-IElementXD-Interfaces-from-Geometry_oM`
on BHoM, BHoM_Engine and other dependent ones

try opening the test file

observe that it does not work

#### Test file(s):
Flip in this folder is a well tested bet on not working, but any of the scripts with errors in the begining does not work, and those without do.
https://burohappold.sharepoint.com/:f:/s/BHoM/ErCbq8ntwdJKi1dnjOUv2ZwBLyBl9mRYRvSv3vSORMbsug?e=YvMu8q

FYI @pawelbaran @IsakNaslundBh

IsakNaslundBh:

Closes #210 

After more digging, the issue did not seem to have anything to do with GH crashes on startup, but had to do with versioning toolkit not being triggered at all for the old way of serialising components. Hence, solution below.

Adding a call to Versioning_Engine, `ToNewVersion` for old component if the first deserialisation of the method returns null.

Code looks safe on the Versioning_Engine side, but needs to be tested through properly to make sure it does not break other things, and can only be tested on scripts saved before new serialisation. This is before early July 2019.

### Test files

Example test files found here: https://burohappold.sharepoint.com/:f:/s/BHoM/ErCbq8ntwdJKi1dnjOUv2ZwBLyBl9mRYRvSv3vSORMbsug?e=YvMu8q

### Additional comments

As this is touching very core framework methods, would be good to get @adecler s signoff before merging.

Problem found while checking versioning after changes in https://github.com/BHoM/BHoM/pull/704

### DIFF
--- a/BHoM_UI/Templates/Caller.cs
+++ b/BHoM_UI/Templates/Caller.cs
@@ -258,9 +258,24 @@ namespace BH.UI.Templates
                     {
                         json = Engine.Versioning.Convert.ToNewVersion(json);
                         obj = BH.Engine.Serialiser.Convert.FromJson(json);
+
+                        //If component was sucessfully upgraded, show the message to inform the user.
+                        //This will happen if: Was not serialised as CustomObject (current serialisation method), failed to de-serialised orginially and was able to be upgraded.
+                        if (obj != null && !m_upgradeMessageShown)
+                        {
+                            MessageBox.Show("BHoM 2.3 or earlier components found in this script." + Environment.NewLine +
+                                            "Automatic upgrade of components to 3.0 has been applied. Some wires may have been disconnected." + Environment.NewLine + Environment.NewLine +
+                                            "To preserve wire connectivity as well you could try upgrading to 2.4 first:" + Environment.NewLine +
+                                            "1. Close the script without saving" + Environment.NewLine +
+                                            "2. Downgrade to BHoM 2.4.beta" + Environment.NewLine +
+                                            "3. Open and save the script" + Environment.NewLine + 
+                                            "4. Upgrade back to current version of BHoM and open the script again.","BHoM auto versioning");
+                            m_upgradeMessageShown = true;
+                        }
                     }
 
                     SetItem(obj);
+
                     if (SelectedItem != null)
                         ItemSelected?.Invoke(this, SelectedItem);
                     return true;
@@ -492,6 +507,7 @@ namespace BH.UI.Templates
 
         protected List<Func<DataAccessor, object>> m_CompiledGetters = new List<Func<DataAccessor, object>>();
         protected List<Func<DataAccessor, object, bool>> m_CompiledSetters = new List<Func<DataAccessor, object, bool>>();
+        private static bool m_upgradeMessageShown = false;
 
         /*************************************/
     }
--- a/BHoM_UI/Templates/Caller.cs
+++ b/BHoM_UI/Templates/Caller.cs
@@ -253,6 +253,9 @@ namespace BH.UI.Templates
                 CustomObject component = obj as CustomObject; // Old component, serialised only with the SelectedItem as object
                 if (component == null)
                 {
+                    if (obj == null)
+                        json = Engine.Versioning.Convert.ToNewVersion(json);
+
                     SetItem(BH.Engine.Serialiser.Convert.FromJson(json));
                     if (SelectedItem != null)
                         ItemSelected?.Invoke(this, SelectedItem);
--- a/BHoM_UI/Templates/Caller.cs
+++ b/BHoM_UI/Templates/Caller.cs
@@ -253,10 +253,14 @@ namespace BH.UI.Templates
                 CustomObject component = obj as CustomObject; // Old component, serialised only with the SelectedItem as object
                 if (component == null)
                 {
+                    //If component failed to de-serialise, try upgrade version.
                     if (obj == null)
+                    {
                         json = Engine.Versioning.Convert.ToNewVersion(json);
+                        obj = BH.Engine.Serialiser.Convert.FromJson(json);
+                    }
 
-                    SetItem(BH.Engine.Serialiser.Convert.FromJson(json));
+                    SetItem(obj);
                     if (SelectedItem != null)
                         ItemSelected?.Invoke(this, SelectedItem);
                     return true;
--- a/BHoM_UI/Templates/Caller.cs
+++ b/BHoM_UI/Templates/Caller.cs
@@ -268,8 +268,9 @@ namespace BH.UI.Templates
                                             "To preserve wire connectivity as well you could try upgrading to 2.4 first:" + Environment.NewLine +
                                             "1. Close the script without saving" + Environment.NewLine +
                                             "2. Downgrade to BHoM 2.4.beta" + Environment.NewLine +
-                                            "3. Open and save the script" + Environment.NewLine + 
-                                            "4. Upgrade back to current version of BHoM and open the script again.","BHoM auto versioning");
+                                            "3. Open and save the script" + Environment.NewLine +
+                                            "4. Upgrade back to current version of BHoM and open the script again." + Environment.NewLine + Environment.NewLine +
+                                            "Note that this message will only be shown once per session of the UI.", "BHoM auto versioning");
                             m_upgradeMessageShown = true;
                         }
                     }


IsakNaslundBh:
Realised my simple change will not work if the method inputs/outputs have been changed in terms of order or count, but only solve the problem of type. WIll have a quick look into solving the other issue as well.

IsakNaslundBh:
Ok, after quite a lot more digging into everything, have a better understanding how things are working, and might understand a bit better why the versioning toolkit was nto called for old scripts in the first place. Some general ideas and thoughts:

- Problem in general only appears for scripts saved before July 2019, when the mode of de-serialisation was changed.
- With the change in this PR all methods that only have their input types changed will be upgraded correctly and work as expected. For example, all changes from the IElements PR will be fixed.
- With the change to this PR, method that had a change in number of inputs will de-serialise correctly, but all wires will be lost. This means all Push/Pull components from scripts before july 2019.
- Without this change, all wires will be kept, but the component will fail to de-serialise. You will then need to add a new component yourself, but be helped with the re-stiching from the old wires.
- For the old way of de-serialising, the method fails to de-serialise (stating the obvious, but...) as itself or the types does not exist. We could still try to parse the json string to get the inputs out, and do some work on them. For example, could get the input could out, as well as types (upgraded) and check if they match the new version, and only allow upgrading if that is the case. As the names of the input params are not stored in the json version of the method, they cant be used to re-stich as is being done atm.

Which leaves a few options that I can think of:
1. Make the change, and live with lost wires (for example all old push/pull components) but get it working for Type upgraders
2. Not make this change at all, live with old scripts failing if methods have changed, but always keep all the wires
3. Add a boolean setting or similar to the versioning toolkit to say that a particular method only allows types to be upgraded, not full methods. Would keep the safe to upgrade cases upgrading, but other cases failing to de-serialise, but keeping the wires
4. Parse the json in BHoM_UI when failing, and as mentioned above, count the inputs and their (upgraded) types, and only upgrade if this still matches. Should give the same result as 3.
5. Some version of 4. Parse the Json inputs, and try to match the types to the new method. If you only have one old type matching one new type, then re-stich, but leave the rest.
6. Make some surgical tweaks to Push and Pull, by overriding Read for them, and check if they were old, and make some hard-coded upgrading for their case only (not liking this to much myself).

Quite a long list. Would be good to get your inputs on this @adecler or if you have a much simpler solution to the whole problem, that I have just missed.

IsakNaslundBh:
To log the conclusion from the call this morning, the suggested way forward after chat between myself, @al-fisher , @pawelbaran , @kThorsager  and @FraserGreenroyd :

- Make the change in this PR, and live with the lost wires, _but_
- Send a pop-up message when it happens, informing the user on what they can do.

What they can do is basically to downgrade their BHoM version to 2.4, as that is the earliest beta making use of the new mode of serialising components. Once saved in 2.4 mode, the new mode of upgrading and re-stitching as well as possible should be working.

al-fisher:
I am testing this now. 
@FraserGreenroyd @pawelbaran @kThorsager  how have you got on?

Also - following https://github.com/BHoM/Dynamo_Toolkit/issues/214 have we confirmed all working smoothly in **Dynamo** (@adecler) and **Excel** (@awakeman) as well as Grasshopper?

pawelbaran:
Works fine on ModelLaundry test files: popup shows + the components deserialize nicely

al-fisher:
> Works fine on ModelLaundry test files: popup shows + the components deserialize nicely

👍 On which UI(s) are you testing @pawelbaran

pawelbaran:
Grasshopper :)

epignatelli:
I have just taken a look at this.
If we are suggesting  to downgrade to 2.4  and then re-upgrading, what is blocking us from automating it?

One other comment I have is that we could create a copy of the current script as it is (like `script.gh.old`, because as soon as you press CTRL + S, we lose the ability to recover it, if I understood correctly?

kThorsager:
Due to how the message is shown through the the BHoM_UI, the message will only appear for the first document you open which encounters the problem, but not subsequent times

BHoM_UI/Templates/Caller.cs
```diff
@@ -253,7 +253,29 @@ public virtual bool Read(string json)
...
+                        //This will happen if: Was not serialised as CustomObject (current serialisation method), failed to de-serialised orginially and was able to be upgraded.
+                        if (obj != null && !m_upgradeMessageShown)
+                        {
+                            MessageBox.Show("BHoM 2.3 or earlier components found in this script." + Environment.NewLine +
+                                            "Automatic upgrade of components to 3.0 has been applied. Some wires may have been disconnected." + Environment.NewLine + Environment.NewLine +
+                                            "To preserve wire connectivity as well you could try upgrading to 2.4 first:" + Environment.NewLine +
+                                            "1. Close the script without saving" + Environment.NewLine +
+                                            "2. Downgrade to BHoM 2.4.beta" + Environment.NewLine +
+                                            "3. Open and save the script" + Environment.NewLine + 
+                                            "4. Upgrade back to current version of BHoM and open the script again.","BHoM auto versioning");
```
FraserGreenroyd: Can we add a space after the comma before the message box header

IsakNaslundBh:
This does not seem to solve the problem for dynamo, upgrading from BHoM 2.3 to current. Pull component gets new inputs, but does not work
IElement components looses all inputs, outputs and wires.

This is tested for Dynamo 1.3:
https://burohappold.sharepoint.com/:f:/s/BHoM/EoW0MLuLCyNCtYMJdjHISYoBy8DJ4bKntBs5guhfK4vihg?e=eZTjNw

Script called: Simple dynamo 1.3 BHoM 2.3 script

Also, I did not even get the pop-up message showing up, so not sure if the read part of the caller is even called.

pawelbaran:
Oke, my test on Dynamo side:
1. Deleted all BHoM assemblies from _appdata_
2. Installed 2.3 from [here](https://burohappold.sharepoint.com/sites/BHoM/Installers/Forms/AllItems.aspx?viewid=9079a548-d015-48f3-842d-253576f13d8c&id=/sites/BHoM/Installers/Betas)
3. Created a script that is available [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FUI%2FIssue211%2DSupportVersioningFrom2%2E3&p=true&originalPath=aHR0cHM6Ly9idXJvaGFwcG9sZC5zaGFyZXBvaW50LmNvbS86Zjovcy9CSG9NL0VvVzBNTHVMQ3lOQ3RZTUpkakhJU1lvQnk4REo0YktudEJzNWd1aGZLNHZpaGc_cnRpbWU9T1U4bFRzQzYxMGc) using Dynamo 1.3, BHoM 2.3
4. Uninstalled BHoM 2.3, recompiled everything from latest branches and... Dynamo does not see BHoM, again. Therefore, so far I can only confirm that Dynamo is hopelessly bad piece of software 🙊 I will plow on trying to resuscitate the zombie - TBC

IsakNaslundBh:
Dynamo:
For me, IElement changes work with the change in this PR, from 2.3 to current.

Push/pull components are not upgraded correctly though. Will try to make a push/pull script from 3.0 and check how that behaves.

pawelbaran:
Unfortunately, it seems that Dynamo will not start working on my machine any more - I need to lean on @IsakNaslundBh, who was lucky enough to have Dynamo cooperating with his machine...

al-fisher:
Thanks guys @pawelbaran @IsakNaslundBh for efforts 🙏

IsakNaslundBh:
Also tested Building a simple adapter script on 3.0. Here it seems the Pull is upgrading fine, although the last wire is plugged in wrong, the component now has 6 inputs instead of 5, but the "activate" input, the last one, still sits an position 5, now action config. After a quick re-plug, the pull worked though.

IsakNaslundBh:
Ok, so, after some more testing of various scenarios my end for Dynamo 1.3. I realised some issues I had with the adapter was coming from robot being frozen for reasons not related to this at all.

### WITHOUT this pr:
#### 2.3 -> 3.1
- Adapter pull/push upgraded correctly, but wires plugged in to wrong position (same index as before)
- IElement components fail to deserialise completely, all inputs, outputs and component itself is lost.

### 2.4 -> 3.1
- Adapter pull/push upgraded correctly, but wires plugged in to wrong position (same index as before)
- IElement upgrades correctly, wires kept and methods run

### WITH this PR:
- Adapter pull/push upgraded correctly, but wires plugged in to wrong position (same index as before)
- IElement upgrades correctly, wires kept and methods run

### 2.4 -> 3.1
- Adapter pull/push upgraded correctly, but wires plugged in to wrong position (same index as before)
- IElement upgrades correctly, wires kept and methods run



### Hence:

From my testing it seems as this PR is not braking anything from the side of dynamo, but is actually helping with de-serialising old dynamo scripts, from BHoM 2.3 using dynamo 1.3.

Would be good if anyone else could confirm they have the same behaviour.

IsakNaslundBh:
@awakeman any luck with the versioning testing on Excel?

kThorsager:
Can confirm that it for the most part works with DynamoSandbox with this PR from 2.3 to 3.1, 
will however add that `Delete` does not upgrade into `Remove` as it should. (Was not sent to Versioning)
What is suprising however is that `Execute` does get upgraded despite not being included in Versioning_Toolkit (Does get sent to Versioning)
image

All this makes me believe that Dynamo relies heavily on the name of the components.

Can also confirm that Versioning is being called and has methods sent through it, so it is not just name matching on the dynamo side, even if that might also happen(?)

Will see if this behaviour (and double check Isak) for the other cases as well

awakeman:
For excel it is as suspected; Since it never made use of `Caller.Read()` or `Caller.Write()` given the way excel doesn't work as components that are serialised but rather formulas that are registered and called, and are thus saved as the formula in the cell rather than a serialised component; it doesn't upgrade components at all. The Dimensional oM change we mostly get away with on account of the name of the IElement* classes being the same and only the name forming part of the registered formula. However `CreateType` formulae are made incompatible by the change in excel (e.g. `=CreateType.Geometry.IElement()` => `=CreateType.Dimensional.IElement()`).

All of this said this PR on the BHoM_UI has no effect, positive or negative, on Excel_UI by virtue of never actually being used.

I am going to raise an issue on Excel for adding support for versioning.

kThorsager:
Continuation of my comment, 
DynamoSandbox 1.3 BHoM 2.4 opening in BHoM 3.1 (with and without this PR).
Here `Excexute` fails properly (as it does not have a upgrade in Versioning)
`Delete` is however still not upgraded to `Remove`, which it should be but it is still not sent to Versioning_Toolkit

And can confirm Isak on Dynamo 1.3 without PR BHoM 2.3 to 3.1

IsakNaslundBh:
Thanks all.

Could I ask you to put an approving review then, so this can be merged @kThorsager @pawelbaran @FraserGreenroyd @awakeman @epignatelli .

 @FraserGreenroyd Have made the chage requested, if you approve if you have tested, or at least dismiss your review if happy with change.

@al-fisher I one more line at the end of the message to state that this message will only be shown once. Good to get your review on the wording.

al-fisher:
> @al-fisher I one more line at the end of the message to state that this message will only be shown once. Good to get your review on the wording.

Looks fine - ta

al-fisher:
🤦‍♂

PR #227, opened 2020-03-19
Ensuring immutability for SetProperty by deep cloning object

### LINKED ISSUE #226: BHoM_UI: SetPropertyValue is not immutable when tweaking inner objects.
#### Description:

When changing an inner property, such as bar.StartNode.Name, the input item is changed as well as the output. this causes weird behaviours in visual programming setups. Will be even more problematic if the values are scaled, as re-running the script then will keep on scaling.

#### Steps to reproduce:

Set a inner property of an object, then check the values of the upstream object.

#### Expected behaviour:

For only the output objects to be changed.

#### Test file(s):

https://burohappold.sharepoint.com/:f:/s/BHoM/Es1XVgVIo6xAqXPRZyndtbkBv89NqlXKMUkRutdRaWHNSw?e=zpHf64

IsakNaslundBh:

Closes #226 

Deep clone input to have its property set to make sure the input does not get changed.

This was causing issues when multiple layer property set was being used, causing the upstream obejcts to be changed.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/Es1XVgVIo6xAqXPRZyndtbkBv89NqlXKMUkRutdRaWHNSw?e=zpHf64

### DIFF
--- a/BHoM_UI/Components/Engine/SetProperty.cs
+++ b/BHoM_UI/Components/Engine/SetProperty.cs
@@ -26,6 +26,7 @@ using BH.UI.Templates;
 using System;
 using System.ComponentModel;
 using System.Reflection;
+using BH.Engine.Base;
 
 namespace BH.UI.Components
 {
@@ -108,6 +109,16 @@ namespace BH.UI.Components
                 base.CompileInputGetters();
         }
 
+        /*************************************/
+
+        public override object Run(object[] inputs)
+        {
+            if (inputs != null && inputs.Length >= 1 && inputs[0] != null)
+            {
+                inputs[0] = inputs[0].DeepClone();
+            }
+            return base.Run(inputs);
+        }
 
         /*************************************/
         /**** Private Fields              ****/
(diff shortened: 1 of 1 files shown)


BHoM_UI/Components/Engine/SetProperty.cs
```diff
@@ -108,6 +109,16 @@ protected override void CompileInputGetters()
                 base.CompileInputGetters();
         }
 
+        /*************************************/
+
+        public override object Run(object[] inputs)
+        {
+            if (inputs != null && inputs.Length >= 1 && inputs[0] != null)
+            {
+                inputs[0] = inputs[0].DeepClone();
```
alelom: Maybe add a comment
```cs
// Deepclone must be done before the properties are set to ensure immutability
```
and something else to explain why it has to be done at the UI level instead than in the Engine perhaps?

al-fisher: Yeah - good idea

adecler: LGTM. Agreed with @alelom

IsakNaslundBh:
@alelom comment added if you could re-review

PR #231, opened 2020-03-26
Avoid generating constructor for types with defined constructor

### LINKED ISSUE #230: BHoM_UI: Avoid generation auto initialisers for certain types?
Raised as question as we are not currently sure about the situation here.

There is this issue happening for the default initialiser for `FragmentSet`:

image

if you select it, you get a full Rhino crash.

This actually prompts a question: do we want to expose that constructor for `FragmentSet` (which is automatically generated)? 

I would probably say no, since it (seems to me like it) means exposing _all available properties from all base classes constructors_, which I don't think makes sense in this case: here our BHoM class inherits (inevitably) `System.Collections.ObjectModel.KeyedCollection`.

The auto-initialiser then includes also the parameters from `System.Collections.ObjectModel.KeyedCollection`:
- `IEqualityComparer<TKey> Comparer`
- `int dictionaryCreationThreshold`

together with a third parameter, that comes from our `BH.oM.Base`:
- `IBHoMFragment fragment`

which does not make sense, and might well be the cause of the full crash.

Now honestly I didn't expect this to happen – I didn't fully understand the range/mechanism of our constructor generator. But certainly this is not an initialiser we want to expose to users.

Maybe an attribute to prevent the generation of an auto-initialiser would be the best solution here?

adecler:

Closes #230 

### Test files
Simply make sure that auto constructors are not available in the UI for types that already contain a constructor (e.g. `IImmutable` and `FragmentSet`). All types without defined constructor should still have one generated.

### Additional comments
- There was no way to detect that a constructor was generated by the compiler. I though I could get that from the `CompilerGeneratedAttribute` attribute but that doesn't work since that attribute is not found even on compiler generated constructors.  So I considered that all constructor without arguments is compiler generated. 
- I removed the check on `Immutable` since the check on the number of non-empty constructor covers them.

### DIFF
--- a/UI_Engine/Query/Items.cs
+++ b/UI_Engine/Query/Items.cs
@@ -119,7 +119,8 @@ namespace BH.Engine.UI
         public static IEnumerable<Type> ConstructableTypeItems()
         {
             return Engine.Reflection.Query.BHoMTypeList()
-                .Where(x => !x.IsNotImplemented() && !x.IsDeprecated() && x?.GetInterface("IImmutable") == null && !x.IsEnum && !x.IsAbstract);
+                .Where(x => x != null && !x.IsNotImplemented() && !x.IsDeprecated() && !x.IsEnum && !x.IsAbstract)
+                .Where(x => x.GetConstructors().Where(c => c.GetParameters().Count() > 0).Count() == 0);
         }
 
         /***************************************************/


al-fisher:
@adecler @alelom just FYI - we have still not closed 3.1 beta release yet. Will happen 13:00 UTC today hopefully. 
So should not be merging new PRs (unless part of testing and wanting to form part of this release). 
This might be fine and so not major issue - but please hold off if planning anything more. Cheers
@FraserGreenroyd FYI

adecler:
Ha, good to know. Will hold off for now then.

PR #244, opened 2020-04-21
Fix wire disconnects for edge cases of Versioning

### LINKED ISSUE #214: BHoM_UI: Disconnects for edge cases of Versioning_Toolkit
#### Description:

any component which the outputs name has been changed since its creation and after which is upgraded by the Versioning_Toolkit gets its output disconnected

#### Steps to reproduce:

1. Create a component
2. Change the output's name in the code with a Reflection Attribute
3. Save the script again, (changing the output does not update the component in the script but still works fine)
4. Have that component be upgraded by the Versioning_Toolkit
5. The output disconnects from whatever it was attached to.

#### Test file(s):

After the migration of `IElement`
https://burohappold.sharepoint.com/:u:/s/BHoM/EVxEhCrhhWJOgo6GIht_7X4Buo0roQUlk-4pF6xEefIkew?e=rlPVIf

adecler:

Closes #214

### Test files
@kThorsager , I'll let you use whatever file works for you (the one provided in the issue was not reproducing the bug for me so I created my own fake scenario).

### DIFF
--- a/BHoM_UI/Templates/Caller.cs
+++ b/BHoM_UI/Templates/Caller.cs
@@ -539,6 +539,8 @@ namespace BH.UI.Templates
             {
                 ParamInfo parameter = newList[i];
                 int oldIndex = oldList.FindIndex(x => x.Name == parameter.Name);
+                if (oldIndex < 0)
+                    oldIndex = i;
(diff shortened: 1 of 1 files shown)


adecler:
Yes, I am relying on names as the only thing that gives a way to find the relation between old parameters and new ones. So I prefer not to change them, especially given that it doesn't break anything to leave them be.

PR #246, opened 2020-04-22
Do not skip `External` keyword in neither namespaces or classes

### LINKED ISSUE #223: BHoM_UI: External subnamespace is explicitly culled out from method tree for components
#### Issue

Thanks to the introduction to the `External` component, an exception for this word in the method tree was added to the Organise method for menu for all engine components.

https://github.com/BHoM/BHoM_UI/blob/3ba297b14facb1c851b4582f55a205380c3c9511/UI_Engine/Compute/Organise.cs#L59-L73

For https://github.com/BHoM/BHoM_Adapter/issues/213 it is critical that this does not happen, as it would mean no effect in the UI, which is one of the main reason that is actioned.

We hence need to remove the "External" word from the `toSkip` list.

@epignatelli would you be able to check what this does to the External component, making sure it still runs as expected, or if we need to add some exception to the rule here?

@al-fisher @adecler @FraserGreenroyd @alelom

epignatelli:

Closes #223

Strategy here is to remove the `External` component, which is unusable in any case since the menu is too slow to display, and only allow the reflected methods in the global menu.

### Test files

This has to tested in tandem with https://github.com/BHoM/BHoM_Adapter/issues/210
You can use https://github.com/BHoM/MidasCivil_Toolkit/pull/203 as a codebase for tests.

### DIFF
--- a/UI_Engine/Compute/Organise.cs
+++ b/UI_Engine/Compute/Organise.cs
@@ -97,7 +97,7 @@ namespace BH.Engine.UI
             List<SearchItem> list = paths.Zip(members, (k, v) => new SearchItem { Text = k, Item = v }).ToList();
 
             //Create method tree
-            List<string> toSkip = new List<string> { "Compute", "Convert", "Create", "External", "Modify", "Query" };
+            List<string> toSkip = new List<string> { "Compute", "Convert", "Create", "Modify", "Query" };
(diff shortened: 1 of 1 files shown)


epignatelli:
Closing this since https://github.com/BHoM/BHoM_Adapter/issues/213#issuecomment-628574216 resorted to using another name.

PR #248, opened 2020-04-25
Improve handling of parameters when upgrading a component

### LINKED ISSUE #242: BHoM_UI: Versioning Methods Not Read Properly
#### Description:

Some methods which have been through Versioning gets a problem in UI.
Specifically ones were the first input is a collection.

```
1. This component failed to run properly.
- Error: Index was outside the bounds of the array.
- Occured in  BH.UI.Templates.MethodCaller.Run(Object[] inputs)
     called from BH.UI.Templates.Caller.Run()
- Are you sure you have the correct type of inputs? Check their description for more details.
```

recreating the method works fine, but is quite annoying

#### Test file(s):

some file with the issue, haven't dared to save since I saw the issue so there's some other things as well
https://burohappold.sharepoint.com/:f:/s/BHoM/EoFRz8dFG9lApm_MB5k0Y_MBnaunhZP6r9oBM6qXEHcyyw?e=CWyAVM

adecler:

Closes #242

See comment https://github.com/BHoM/BHoM_UI/issues/242#issuecomment-619346538

### Additional comments
- @epignatelli , I added you as a reviewer to make sure that I didn't miss anything by changing your original code.
- There is clearly a need to properly list all the possible case of what changes can happen to inputs.outputs and make sure we have a strategy for each. This PR is mainly to patch things up in the mean time.

### DIFF
--- a/BHoM_UI/Templates/Caller.cs
+++ b/BHoM_UI/Templates/Caller.cs
@@ -536,11 +536,58 @@ namespace BH.UI.Templates
 
         protected void FindOldIndex(List<ParamInfo> newList, List<ParamInfo> oldList)
         {
+            List<int> newToMatch = Enumerable.Range(0, newList.Count).ToList();
+            List<int> oldToMatch = Enumerable.Range(0, oldList.Count).ToList();
+
+            // First match using names
             for (int i = 0; i < newList.Count; i++)
             {
                 ParamInfo parameter = newList[i];
                 int oldIndex = oldList.FindIndex(x => x.Name == parameter.Name);
+                if (oldIndex >= 0)
+                {
+                    newToMatch.Remove(i);
+                    oldToMatch.Remove(oldIndex);
+                }
                 parameter.Fragments.AddOrReplace(new ParamOldIndexFragment { OldIndex = oldIndex });
+            }
+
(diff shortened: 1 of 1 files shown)


adecler:
The second commit should 
Fix #214

IsakNaslundBh:
@awakeman does this in any way affect your work in https://github.com/BHoM/Excel_Toolkit/pull/208

kThorsager:
This fixed #214 and #242 is better, but I'm getting other bugs now.

Namely inputs of type `List<IElement2D>` think they are `List<IElement1D>` (last file) Attempting to use IElement1Ds in those methods does not work, Error: unable to cast from `List<Bar>` to `List<IElement1D>`

https://burohappold.sharepoint.com/:f:/s/BHoM/EoFRz8dFG9lApm_MB5k0Y_MBnaunhZP6r9oBM6qXEHcyyw?e=iTmhlx

adecler:
Thanks @kThorsager , good catch !
This actually doesn't have anything to do with this PR, this was just hidden by other problems that are now solved.
I have raised the issue in the reflection engine : https://github.com/BHoM/BHoM_Engine/issues/1731
and raised the PR to fix it: https://github.com/BHoM/BHoM_Engine/pull/1732

Given that this was the last problem with this PR, I think we can safely merge.

PR #259, opened 2020-05-08
UI_PostBuild: Fix UnauthorizedAccessException on PostBuild CopyUpgrades

### LINKED ISSUE #251: Throwing System.UnauthorizedAccessException on PostBuild CopyUpgrades
I am getting an error from the post build events when building BHoM UI.

`Error		The command "call "C:\BHoMGitHub\BHoM_UI\Build\UI_PostBuild.exe" ..\..\ "C:\Users\afisher\AppData\Roaming\BHoM\Assemblies"" exited with code -532462766.`	

This is failing as part of the `CopyUpgraders` and on investigation is on line 65 when attempting to access a folder which it does not have permissions for.	Specifically a `System.UnathorizedAccessException`

https://github.com/BHoM/BHoM_UI/blob/35fd6b747cea05582d04594e19f44e30e98ce876/UI_PostBuild/CopyUpgrades.cs#L65

As discussed with @IsakNaslundBh a few options to consider here to catch this and make fail safe.

- First would be to have a parent for-loop iterating across each toolkit folder - before calling `Directory.GetFiles`. This would allow as to gracefully catch an exception for individual toolkits/folders - without crashing out entirely and losing abitlity to copy any of the upgraders.

- The second option - which I think might be a preference - and perhaps in addition to option one, is to limit the search to top level of toolkit folders etc. - i.e. not fully recursive. And thus also being more deliberate about aggregating files only from where they should be expected.

Naturally this second option constrains more where upgraders can be stored, to a more defined BHoM standard upgrader compliance - (both good and bad) but will equally reduce risk further of mining unexpected files, access to miscellaneous folders etc. 
As code base continues to grow. No bad thing I think to enforce a relatively low threshold compliance for file locations.

@adecler good to get your thoughts.
Thanks again @IsakNaslundBh
@rolyhudson @kThorsager this was the issue discussed this morning in our stand up FYI

adecler:

Closes #251

I decided to go with Al'second option. I mentioned that it would not work for the Dynamo folder but I started to think this is actually a good thing. That folder is purposefully structured to avoid being transferred to the Roaming folder so it makes sense that it also avoid the versioning picking. Something to deal with Dynamo toolkit either way and should not influence what is done here.

### DIFF
--- a/UI_PostBuild/CopyUpgrades.cs
+++ b/UI_PostBuild/CopyUpgrades.cs
@@ -62,7 +62,7 @@ namespace BHoM_UI
             BsonDocument versioning = CreateEmptyUpgradeDocument();
 
             // Collect Versioning files
-            string[] versioningFiles = Directory.GetFiles(sourceFolder, versionFileName, SearchOption.AllDirectories);
+            List<string> versioningFiles = GetVersionFiles(sourceFolder, versionFileName);
             foreach (string file in versioningFiles)
                 ReadVersioningFile(file, versioning);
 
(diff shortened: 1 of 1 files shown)


IsakNaslundBh:
@al-fisher do you want to have a first look at this, as I did not experience any problems before.

PR #263, opened 2020-05-14
Align with auto-versioning during serialisation

### LINKED ISSUE #262: BHoM_UI: Align with auto-versioning during serialisation
#### Description:
The components cannot detect that they have been upgraded by just having a failed serialisation anymore (since upgrade will happen automatically). This means that, in the current state, component will not detect tha thy have been updated and will not fix their inputs/outputs and connections.

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_Engine/pull/1762  

(at least on a behavioural level so recommend to review them together)

   
Closes #262

Without this, you will have missing wires when testing https://github.com/BHoM/BHoM_Engine/pull/1762  

### Test files
Old files that contain components that need upgrading

### DIFF
--- a/BHoM_UI/BHoM_UI.csproj
+++ b/BHoM_UI/BHoM_UI.csproj
@@ -67,9 +67,6 @@
       <HintPath>..\..\BHoM_Engine\Build\Library_Engine.dll</HintPath>
       <Private>False</Private>
     </Reference>
-    <Reference Include="MongoDB.Bson, Version=2.4.4.0, Culture=neutral, processorArchitecture=MSIL">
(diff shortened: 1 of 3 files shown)


adecler:
Now that there is no dependency left, I have remove the reference to Mongo Bson

adecler:
> Only thing is no message is shown any longer for older scripts being upgraded with a big risk of loosing their wires. But will leave it to you @adecler and @al-fisher to say if that is a big issue or not.

Unfortunately that was the price to pay for having the auto-versioning happening during deserialisation. The UI is not aware that versioning happened anymore so doesn't know when to trigger the warning. Thankfully, this is only for the older scripts that didn't save the inputs/outputs.

PR #271, opened 2020-05-26
BHoM_UI: Add support for "&" ByRef types in Caller

### LINKED ISSUE #270: Add support for "&" ByRef types in Caller
#### Description:

When [using Speckle_Toolkit from within Excel](https://github.com/BHoM/Speckle_Toolkit/pull/91) I'm getting an error when Pushing. This error originates in `Template\Caller.cs`, line 480, and is:

```
Cannot convert from System.Double& to System.Double
```

the ampersand types are byRef types. With those types, an additional method call is needed to get the effective type:
```cs
if (t.IsByRef)
    t = t.GetElementType();
```

This can be added in.

alelom:

Closes #270

### DIFF
--- a/BHoM_UI/Templates/Caller.cs
+++ b/BHoM_UI/Templates/Caller.cs
@@ -477,7 +477,7 @@ namespace BH.UI.Templates
         {
             UnderlyingType subType = dataType.UnderlyingType();
             string methodName = (subType.Depth == 0) ? "SetDataItem" : (subType.Depth == 1) ? "SetDataList" : "SetDataTree";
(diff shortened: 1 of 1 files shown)


adecler:
I don't see a problem with the code itself although I haven't tested it myself.
I would just flag up that this probably means we only have partial support for ByRef types within the UI. Since partial support tends to create confusing problems, I would recommend raising a separate issue as a minimum if this is too much work to take care of here.

BingWangUS:
This is a great improvement for searching. Cannot have "space" is kind of a paint before this.

PR #275, opened 2020-06-03
Update postbuild and references to ProgramData

adecler:
https://github.com/BHoM/admin/issues/5

- Add two new arguments to UI_PostBuild: 
  - `--onlyUpgrades`: use this to make sure only upgrade files are created (no dll or dataset)
  - `--cleanTargetDir`: use this to have the target folder cleared first
- Keep the content of the existing `Upgrades.json` file by adding to it instead of replacing it
- Change BHoM_UI postBuild target folder

### Additional comments
- @IsakNaslundBh , I kept the target folder to `Assemblies/bin` to mirror what you did in the engine (`ToNewVersion` file) but actually very easy to use the `BHoM/Upgrades` folder instead as nothing changes in the code, only in the post build command 
- To test the Post_Build process properly, please use https://github.com/BHoM/Versioning_Toolkit/pull/77

### DIFF
--- a/BHoM_UI/BHoM_UI.csproj
+++ b/BHoM_UI/BHoM_UI.csproj
@@ -233,7 +233,7 @@
   <PropertyGroup>
     <PostBuildEvent>call "$(TargetDir)UI_PostBuild.exe" ..\..\ "C:\ProgramData\BHoM\Assemblies" --onlyUpgrades
 
-xcopy \"$(TargetDir)$(TargetFileName)\"  \"C:\\ProgramData\\BHoM\\Assemblies\" /Y</PostBuildEvent>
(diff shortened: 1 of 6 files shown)


FraserGreenroyd:
Ok got Core to fail on Excel now rather than Grasshopper so that's better for now given Excel has no PR yet.

al-fisher:
Thanks @adecler - as just discussed with @IsakNaslundBh we also like the `BHoM/Upgrades` as a name and as a clear clean location
@IsakNaslundBh is to make tweaks necessarily as highlighted

IsakNaslundBh:
Have updated the Engine Branch with changes. Assumed the the bin folder would not be needed any longer as well, now that in a separate folder, but that we would have `BHoM/Upgrades/BHoMUpgrader32/BHoMUpgrader32.exe` for example

PR #285, opened 2020-06-29
Set m_OriginalMethod to null when selecting a type

### LINKED ISSUE #284: BHoM_UI: m_OriginalMethod is not nulled out when type is selected for CreateObject
#### Description:

When `SetItem` is called for the general case that a method is being picked, `m_OriginalMethod` is being handled and set to the corresponding method or nulled out if not used in the `SetItem` on the method called.
On the CreateCaller on the other hand, when a type is selected, `m_OriginalMethod` is not touched.

For Grasshopper and dynamo this is not an issue, as they have a one-to-one relation between a component and one caller. For excel on the other hand this becomes an issue, as Excel_UI only owns one of each caller type, which is being used for all the calls that goes through that particular caller.

This means that if a create method is called, followed by a property assignment method, basically method followed by type, `m_OriginalMethod` is still set to the method.

This in turn causes issues with the serialisation of the component, as selected item prioritises `m_OriginalMethod` over `SelectedItem` when turned to JSON.

#### Steps to reproduce:

In excel, use a create method followed by a property assignment method and look how they are being serialised. 

#### Expected behaviour:

For `m_OriginalMethod` only to be set when using a method and for the component to de-serialise correctly.

#### Test file(s):

This is all happening on adding methods, hence no test file can be provided. Try Creating an object with a create method, then creating an obejct with a property initialiser. Then either look at how the methods have been stored, and/or save and reopen the sheet.

IsakNaslundBh:

Closes #284

as stated in issue:

When `SetItem` is called for the general case that a method is being picked, `m_OriginalMethod` is being handled and set to the corresponding method or nulled out if not used in the `SetItem` on the method called.
On the CreateCaller on the other hand, when a type is selected, `m_OriginalMethod` is not touched.

For Grasshopper and dynamo this is not an issue, as they have a one-to-one relation between a component and one caller. For excel on the other hand this becomes an issue, as Excel_UI only owns one of each caller type, which is being used for all the calls that goes through that particular caller.

This means that if a create method is called, followed by a property assignment method, basically method followed by type, `m_OriginalMethod` is still set to the method.

This in turn causes issues with the serialisation of the component, as selected item prioritises `m_OriginalMethod` over `SelectedItem` when turned to JSON.

### Test files

This is all happening on adding methods, hence no test file can be provided. **In Excel:** Try Creating an object with a create method, then creating an obejct with a property initialiser. Then either look at how the methods have been stored, and/or save and reopen the sheet.

Alsno need to make sure this does not cause any issues elsewhere in the other UIs.

### DIFF
--- a/BHoM_UI/Components/oM/CreateObject.cs
+++ b/BHoM_UI/Components/oM/CreateObject.cs
@@ -128,6 +128,7 @@ namespace BH.UI.Components
             else if (item is Type)
             {
                 SelectedItem = item;
+                m_OriginalMethod = null;
                 Type type = item as Type;
                 Name = type.Name;
                 Description = type.Description();


pawelbaran:
Seems to be fine with Dynamo too.

PR #290, opened 2020-07-10
Adding ability to log component usage

### LINKED ISSUE #289: Create log files recording usage of components
#### Description:
To better understand which part of the BHoM is use, record the usage of each component into a log file

adecler:

Closes #289 

This saves the component usage in log files. The files are located in `C:\ProgramData\BHoM\Logs`. 

- To avoid the potential issue of multiple UIs writing to the same file, I simply have each UI writing into its own. The fil name convention is `Usage_<uiName>_<CreationTimeInTicks>.log`. I think that reduces teh risk of having files with the same name down to zero :smile:.
- Each record saves the following:
  - Time of recording
  - UI hosting the component
  - Component Id
  - Selected BHoM item for the component
  - Error events (that should help us identify which component needs fixing even if the user doesn't raise an issue)
- Note that I considered recording the name of the file containing the component but decided against it because this gives away information that people might not like to share (like project name). 
- There is only one record for each run of teh component even if GH is running it once per input (solved in the [related GH PR](https://github.com/BHoM/Grasshopper_Toolkit/pull/526))
- Multiple runs of a components are however saved as separate records to make sure everything is recorded even if the UI crashes (i.e. saving directly to file instead of temporary in memory). The 'summarisation' of the records will be done when sending to the database.

### Test files
- Don't forget to also compile [the PR for the corresponding UI](https://github.com/BHoM/Grasshopper_Toolkit/pull/526) (I will do Excel and Dynamo after merging this PR to avoid mirroring potential changes in a lot of PRs. I hope you don't mind testing with GH :smile:).
- Just open any GH file and make sure a log is created for it. You can already read that log while GH is still running but you cannot delete it until Rhino is closed.

### DIFF
--- a/UI_Engine/UI_Engine.csproj
+++ b/UI_Engine/UI_Engine.csproj
@@ -64,9 +64,9 @@
       <HintPath>C:\ProgramData\BHoM\Assemblies\Reflection_oM.dll</HintPath>
       <Private>False</Private>
     </Reference>
-    <Reference Include="Serialiser_Engine, Version=3.0.0.0, Culture=neutral, processorArchitecture=MSIL">
+    <Reference Include="Serialiser_Engine">
       <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\..\..\..\..\..\..\ProgramData\BHoM\Assemblies\Serialiser_Engine.dll</HintPath>
+      <HintPath>C:\ProgramData\BHoM\Assemblies\Serialiser_Engine.dll</HintPath>
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
--- a/UI_Engine/UI_Engine.csproj
+++ b/UI_Engine/UI_Engine.csproj
@@ -64,6 +64,10 @@
       <HintPath>C:\ProgramData\BHoM\Assemblies\Reflection_oM.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="Serialiser_Engine, Version=3.0.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\..\..\..\..\..\ProgramData\BHoM\Assemblies\Serialiser_Engine.dll</HintPath>
+    </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Xml.Linq" />
@@ -74,6 +78,7 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="Compute\LogUsage.cs" />
     <Compile Include="Compute\LoadAssemblies.cs" />
     <Compile Include="Compute\Organise.cs" />
     <Compile Include="Convert\FromProperty.cs" />
--- a/UI_Engine/Compute/LogUsage.cs
+++ b/UI_Engine/Compute/LogUsage.cs
@@ -1,0 +1,108 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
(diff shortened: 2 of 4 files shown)


UI_Engine/UI_Engine.csproj
```diff
@@ -64,6 +64,10 @@
       <HintPath>C:\ProgramData\BHoM\Assemblies\Reflection_oM.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="Serialiser_Engine, Version=3.0.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\..\..\..\..\..\ProgramData\BHoM\Assemblies\Serialiser_Engine.dll</HintPath>
```
FraserGreenroyd, line 69: Reference should be set to `ProgramData` - probably explains why CheckCore is failing.

adecler: Now fixed

adecler:
Tanks @FraserGreenroyd , 

I cannot really see how the slow down of the wire menu can be due to this PR as it is only impacting the `SolveInstance` method of a component. I have placed breakpoints where the code was changed and tested the wire menu just to be sure. None of the breakpoints where hit during the menu phase (only hit AFTER the component is created if it can be ran without requiring inputs from the user). SO I think we can safely say this PR is fine on that level.

Also worth noting your comment on the multiplication of files. The idea is that the BHoM_analytics toolkit will delete them when they have been processed BUT it is true that will stay forever if the toolkit is not installed. We can add a system to this PR that delete files older than a given period of time. @al-fisher , what do you think ? It could be added to this PR so I am not merging yet.

al-fisher:
> We can add a system to this PR that delete files older than a given period of time. @al-fisher , what do you think ? It could be added to this PR so I am not merging yet.

Yes - I think we need a strategy here and plan as part of this PR as we are creating a lot of data - which can be annoying for users to have to purge themselves. In alpha - but worth planning now before merge I would agree

al-fisher:
Then on testing this now, I was too initially thinking we might want a single file per UI. However, if not purging for a while, it might be problematic to open and append growing files, rather than simply creating new per session. So think fine.

Have a couple of comments about naming and clarity to the users:

Is it worth naming the files based on more understandable UTC date and time (inc. millisecs) - rather than just ticks? 
What do you think? People will undoubtedly look at (and in) a log file - and so I think a code that can be more meaningful can help?

Similar to the above - interested in whether useful to consider making a Log Item object already now - rather than storing as Custom. Might there be benefits down the line - for both manipulation and versioning? @adecler what do you think?
On this one - I suppose might be a balance with storing maximum information with minimum data possible. So custom might be preferable. Good to hear thoughts

adecler:
Thanks @al-fisher ,

I have added to removal of log files older than a week at start. Happy to go for a different period but that felt reasonable.

> Is it worth naming the files based on more understandable UTC date and time (inc. millisecs) - rather than just ticks?
What do you think? People will undoubtedly look at (and in) a log file - and so I think a code that can be more meaningful can help?

If you are referring to the filename, the creation time of the file is provided directly by the file explorer (at least if you are in details view mode). This is why I decided to go with a more machine friendly representation (sooo much easier to process in the code) as the default string version of a data contains non-valid characters for file names.

If you are referring to the way dates are stored in Json, I am afraid there is little I can do about that as it is enforced by Mongo Bson. Well, we could always override it but that means making it incompatible with all serialised info we have.

Also keep in mind that, when people have BHoM_Analytics installed, those files will be removed as soon as their data is sent to the central database anyway. So their format should not really matter as they will have a very short live.

> interested in whether useful to consider making a Log Item object already now - rather than storing as Custom

That was my original thinking but I didn't want to sort out the issue of dependencies. Basically, the BHoM_UI cannot depend on BHoM_Analytics as it is not guaranteed to be installed. BHoM_Analytics could depend on BHoM_UI but somehow it felt wrong. The remaining option would a third party repo like `BHoM` but it didn't have any section appropriate for it to store that info and didn't want to spread this PR more than I already did. 

Edit: Actually, I decide that having the object stored in UI_oM wasn't really a problem after all. So I just made the commit for that.

al-fisher:
> Edit: Actually, I decide that having the object stored in UI_oM wasn't really a problem after all. So I just made the commit for that.

Like the conclusion :smile: - cheers. Will review

PR #292, opened 2020-07-15
New GetInfo component

### LINKED ISSUE #291: Make the GetInfo component more useful
#### Description:
At the moment, the component just return `This is the BHoM`. Not massively useful 🤔

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_Engine/pull/1901

   
Closes #291

Provides a lot more information that the previous `This is the BHoM` string. 

image

One think that I cannot get right now is the version of the last installer (if any). Looking at teh registry, this is all I have:

image

A quick look at the installer's code seems to confirm that the only thing it does with the registry (Except for handling Excel) is to write a `1` in the BHoM registry. It would be great if it could write teh name of the installer instead. 

### Test files
Just create a `GetInfo` component (the `?` in the Engine section for those that are wondering) and enjoy 😉


BHoM_UI/Components/Engine/GetInfo.cs
```diff
@@ -60,11 +63,58 @@ public class GetInfoCaller : MethodCaller
...
+        [MultiOutput(4, "types", "List of objects types available in this version of the BHoM.")]
+        [MultiOutput(5, "methods", "List of methods available in the version of the BHoM.")]
+        [MultiOutput(6, "adapters", "List of adapters available in the version of the BHoM.")]
+        [MultiOutput(7, "website", "BHoM website.")]
+        [MultiOutput(8, "github", "GitHub organisation containing all our open-source toolkits.")]
+        [MultiOutput(9, "wiki", "Need more details on how to use the BHoM or how to contribute? This is the place to go.")]
+        public static Output<string, string, string, List<string>, List<Type>, List<MethodInfo>, List<Type>, string, string, string> GetInfo()
         {
-            return "This is the BHoM";
+            string description = "This the BHoM (Buildings and Habitats object Model), a collaborative computational development project for the built environment. "
```
FraserGreenroyd, line 79: Should this be `This is the BHoM` rather than `This the BHoM`?

BHoM_UI/Components/Engine/GetInfo.cs
```diff
@@ -60,11 +63,58 @@ public class GetInfoCaller : MethodCaller
...
+                + "The whole BHoM project uses an open-source model for project architecture, co-creation and planning. So explore, experiment and contribute to both the source code and the wiki. "
+                + "Sharing and building our code together in this open-source type approach means we can feed off and pool our disparate knowledge, experience and expertise towards a common goal � better design.\n\n"
+                + "Also, as well as creating a common language of BHoM objects - the large number of repositories contain a variety of different plugins and code to operate on BHoM objects and link the BHoM with our favourite software and tools. "
+                + "Much of the core code is written in C#. But we also have code in JavaScript, C++, Python and visual programming languages such as Grasshopper User Objects and Dynamo Custom Nodes all forming part of the BHoM.";
+
+            string version = Engine.Reflection.Query.BHoMVersion();
+            string installer = "";
```
FraserGreenroyd, line 91: I can't see where this variable is assigned, I'm assuming this is placeholder as it needs the investigation on the Installer to provide it's version somewhere usefully - but just double checking in case this PR was meant to provide more than just an empty string placeholder?

BHoM_UI/Components/Engine/GetInfo.cs
adecler: done, thanks 😄

BHoM_UI/Components/Engine/GetInfo.cs
adecler: Correct. I have added a `TODO` to make this clearer

al-fisher:
FYI - updated the "Depends On" in PR body

BHoM_UI/Components/Engine/GetInfo.cs
```diff
@@ -60,11 +63,58 @@ public class GetInfoCaller : MethodCaller
...
+                + "It is crafted as transdisciplinary, software-agnostic and office/region/country-invariant, and therefore would be nothing without our active community and wide range of contributors.\n\n"
+                + "The whole BHoM project uses an open-source model for project architecture, co-creation and planning. So explore, experiment and contribute to both the source code and the wiki. "
+                + "Sharing and building our code together in this open-source type approach means we can feed off and pool our disparate knowledge, experience and expertise towards a common goal � better design.\n\n"
+                + "Also, as well as creating a common language of BHoM objects - the large number of repositories contain a variety of different plugins and code to operate on BHoM objects and link the BHoM with our favourite software and tools. "
+                + "Much of the core code is written in C#. But we also have code in JavaScript, C++, Python and visual programming languages such as Grasshopper User Objects and Dynamo Custom Nodes all forming part of the BHoM.";
```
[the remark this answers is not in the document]
al-fisher: Hey - thought it would be good to make this quite a bit shorter. Means we don't have to have lots of duplicated text in the code to worry about revising and keeping up to date. As that prose from the wiki could indeed probably do with a refresh anyway. 
Repeated the link to the website - although like the explicit website output too!

adecler:
@al-fisher , I fixed the 3 descriptions and the toolkit output. For the later I decided to
- provide the full dll names so we can see which "discipline" has a oM, an Engine,...
- sort them alphabetically to make that easier to browse and more useful when checking if you have a specific dll
- renamed the output `libraries`. This is consistent the name `dynamically loaded library`, with how dynamo name them in its packages window (funny I finally agree with Autodesk on something), and with how grasshopper names the plugin folder (although GH doesn't distinguish between packages and libraries). 
- added `dll` in parenthesis in the description in case some find library less clear than dll.

PR #293, opened 2020-07-15
Explode to give descriptions of properties

### LINKED ISSUE #191: BHoM_UI: Explode to give descriptions of properties
#### Description:

Similarly to how the Create component gives descriptions for input would be nice if the explode could do the same when the input is a non-custom obeject.

adecler:

Closes #191 

Explode now give a description of the properties **_IF_** there is a single type of objects provided as input. I separated that case of a single type from the others to make it clearer in the code.

### Test files
Just try a few explodes on different objects

### Additional comments
- I am aware that if you do something like explode `Point` and then change the input type to `Vector`, the description is not going to change because they have the exact same list of properties (and same names). This can only be fixed in each UI. Keep in mind that this is very rare since it needs the exact same properties.
- I am not too happy with the increase of code complexity (number of ifs) but I don't think it can be avoided without a much larger refactoring. This is due to Dictionaries and custom objects being exceptional cases.

### DIFF
--- a/BHoM_UI/Components/Engine/Explode.cs
+++ b/BHoM_UI/Components/Engine/Explode.cs
@@ -204,35 +204,19 @@ namespace BH.UI.Components
                 return false;
 
             // Group the objects by type
-            Dictionary<string, List<Type>> properties = new Dictionary<string, List<Type>>();
-
             var groups = objects.Where(x => x != null).GroupBy(x => x.GetType());
-            if (groups.Count() == 0)
-                return false;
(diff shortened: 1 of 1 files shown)


adecler:
Thanks @al-fisher , this is something to be fixed in Grasshopper. 
I have made a PR here: https://github.com/BHoM/Grasshopper_Toolkit/pull/527
Note that this also takes care of the problem I mentioned earlier with Point and Vector.

PR #295, opened 2020-07-17
Create toolkit setting system

adecler:
### NOTE: Depends on 
Supports the [PR in BHoMAnalytics](https://github.com/BHoM/BHoMAnalytics_Toolkit/pull/2).

This now provides a solution for toolkits to save some settings in `C:\ProgramData\BHoM\Settings`. Basically, each toolkit can have its own config file with the corresponding name there. I have exposed the methods to read and save settings so users can modify those settings without having to mess directly with a json file. Two immediate usages of this come to mind:
- BHoMAnalytics can store the information about the database where the usage data is sent -> that's what the BHoMAnalytics PR does)
- User settings for the UI (e.g. user tags, customised look,....) -> soon in oyu favorite BHoM UI 😄 

The other thing I did here is to let the toolkits execute some code when the BHoM is started. This is directly relevant for BHoMAnalytics again since it is now able to send the usage logs to the database without having any hard-coded reference to BHoMAnalytics in the BHoM UI at all. This means that if developers outside the Collective need a similar initialisation phase for their toolkit, they don't need to modify the BHoM UI. 

Here's how it works:
- At launch, BHoM UI will look for all settings in the Settings folder that implement the `IInitialisationSettings` interface. 
- It will then try to find the method provided as a string by the only property of that interface
- If that method is found, it will be run. 
   
See https://github.com/BHoM/BHoMAnalytics_Toolkit/issues/1

### Changelog
- New toolkit settings system with ability to save and load settings to and from `C:\ProgramData\BHoM\Settings`
- Ability for toolkits to Initialise themselves when UI is started if they register an initialisation method in `BHoM\Settings'

### DIFF
--- a/BHoM_UI/BHoM_UI.csproj
+++ b/BHoM_UI/BHoM_UI.csproj
@@ -151,6 +151,7 @@
     <Compile Include="Components\oM\CreateType.cs" />
     <Compile Include="Components\oM\CreateObject.cs" />
     <Compile Include="Components\oM\CreateDictionary.cs" />
+    <Compile Include="Global\Initialisation.cs" />
     <Compile Include="Global\GlobalSearch.cs" />
     <Compile Include="Global\SearchMenu.cs" />
     <Compile Include="Global\SearchMenu_WinForm.cs" />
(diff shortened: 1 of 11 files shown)


adecler:
This is now a slightly bigger PR as  I have added the framework for toolkit permanent settings saved in `C:\ProgramData\BHoM\Settings\`.

image

I've added @IsakNaslundBh  as reviewer as it will concern toolkit owners in general.

adecler:
Ok guys, this is now ready for review. 
Make sure to check the updated version of the Pr information.

PR #297, opened 2020-08-01
BHoM_UI  Framework Refactoring

### LINKED ISSUE #156: CreateObject component: The order of the inputs depends on the sequence we select/deselect them
#### How to replicate:

1. Checkout this branch https://github.com/BHoM/BHoM_UI/pull/152 and compile the code accordingly
1. Open Grasshopper
1. Create a `CreateObject` component
1. Select the `oM.Structure.Elements.Bar` constructor by object initializer
1. Right click, and use the Add/Remove Inputs functionality
1. Remove the `StartNode` property
1. Re-add the `StartNode` property
1. Note that it appears at the bottom now.

#### Expected behaviour:

I think we can use the index of the element in the properties list of the object, in conjunction with the `AddInput` method to place the input correctly. 
```pseudo
if IndexOfProperty < InputParams.Length 
  Add it to the right place (Grasshopper should take care of shifting the rest of them)
  OnBHoMUpdates()
else
   Add it to the end of the list
```

### LINKED ISSUE #176: BHoM_UI: Property assigment creates for Requests does not show up in CreateRequest component
#### Description:

Since enabling property assignment methods in the CreateObejct component, of no more need for generating Create method for requests without any default `Create` methods in the engine.

The CreateRequest needs to be updated to match this behaviour. Right now only Requests with a create method show up there.

#### How to replicate:

Try to create for example a `BarResultRequest` in the CreateRequest component

#### Expected behaviour:

For all proeprty assigment creates for requests to show up in the CreateRequest component

#### Test file(s):

No testfile needed.

### LINKED ISSUE #189: BHoM_UI: Enforce immutability at the level of the UI
### This issue has been archived 

Historic issue archive can be found [here](https://burohappold.sharepoint.com/sites/BHoM/01_Master%20File/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F01_Master%20File%2F999_BHoM_GitHub_History_Archive)

### LINKED ISSUE #281: BHoM_UI: Can't Ctrl+Z "past" a explode component
#### Description:

I placed a explode component on the canvas, then tried to Ctrl+Z to undo it, but the explode component does not disappear and nothing else does either.

#### Steps to reproduce:

Had this setup in Grasshopper, seems to work for any exploded object, but have not tested for other UIs

image

### LINKED ISSUE #282: BHoM_UI: weird stuff showing up in CreateObject -> Geometry
#### Description:

As in the screenshot, seems to be the case for both Grasshopper and Dynamo. It does not look harmful, more of a minor UX glitch.

image

@adecler please feel free to reassign labels or migrate to other repo if that is not relevant to BHoM_UI.

### LINKED ISSUE #296: Handle the case when a list is a property of CustomObject for SetProperty and GetProperty
#### Description:

image

image

### LINKED ISSUE #298: Add CallerName property to UsageLogEntry
#### Description:
This is necessary since some components don't have a selected item (e.g. 'Explode' or 'FromJson')
Will also have to be reflected in BHoMAnalytics toolkit when it summarise the data to group by SelectedItem and caller Name.

Also add UI version/info like Rhino 5 vs 6 or Dynamo 1.3 vs 2.0, Excel version, etc

adecler:
### NOTE: Depends on

https://github.com/BHoM/BHoM_Engine/pull/1924

The exact list of issues addressed in this PR will be defined later as this major refactoring will invalidate quite a few pending issues. Hopefully, it will not create too many new ones 😋 

This PR aims at simplifying the UI code in general and avoiding as much as possible code that is difficult to modify without breaking a lot of things. There is two sides to that simplification:
- The BHoM_UI itself: reduce code complexity, take as much of the complexity load away from the individual UIs
- The code on each UI: code should be as small as possible thanks to the changes on the BHoM_UI itself.

Currently, I have only done the first phase of focusing on the BHoM_UI itself without breaking the individual UIs. Sprint 3 will then focus on the simplification of individual UIs by leveraging teh changes in BHoM_UI.

This PR is just for visibility at this stage. I don't think a review is necessary yet as quite a lot more changes are coming over the next two weeks. I will comment here when I think it is valuable for you to start using this as your main branch.

Fixes #156
Fixes #176
Fixes #189
Fixes #281
Fixes #282
Fixes #296
Fixes #298

### Test files
In progress: (quite a lot more to do)
- [Grasshopper](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/Installers/Scripts/BuroHappold_BHoM_v3.3/0047_Grasshopper%20Toolkit/UI_Testing-Grasshopper.gh?csf=1&web=1&e=r0WguD)

### Changelog
TODO

### Additional comments
- The name of the branch is not linked to a specific issue as it generally should because this PR will invalidate quite a few existing issues. So I thought it would make sense to differentiate it from the rest. I hope it doesn't break anything on the CI side @FraserGreenroyd . If it does, I can always change it.
- I have started experimenting with the idea of passive vs active BHoM objects (active objects being UI callers and adapters, passive being what is in the oM, both type being serialisable) but I think we'll need a proper workshop to refine that in details.


adecler:
I should now be done with the changes required to support the refactoring of Grasshopper. So hopefully, only minor changes from now on.

Note that Excel will be done next week (well, this week if you are reading this on Monday). Changes to Dynamo will be done as soon as possible. The idea is that they are both compiling but this is as much as I can guarantee right now. So I recommend to use this branch only when you are using Grasshopper exclusively.

I will also provide the drawings representing the new code structure as soon as I have taken care of Excel and Dynamo. Those are not essential for you to start using this branch but will definitely be useful if you want to do some debugging. 

For now, please start using this branch as much as possible without worrying too much about reviewing the code.

Thanks guys  :smile:

al-fisher:
> I will also provide the drawings representing the new code structure as soon as I have taken care of Excel and Dynamo. Those are not essential for you to start using this branch but will definitely be useful if you want to do some debugging.

Thanks @adecler 
Switching branches and will run as usual for coming days

BHoM_UI/Components/Adapter/Push.cs
```diff
@@ -25,17 +25,16 @@
...
-using BH.UI.Templates;
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 
-namespace BH.UI.Components
+namespace BH.UI.Base.Components
```
IsakNaslundBh, line 35: Wonder if the namespace change here means that the push method on this class needs a `PreviousVersion` attribute, as versioning from older version of this method points to a class that does not exist any longer?

adecler:
Ok, the bug with the deserialisation of old adapter components was more complicated than I thought:
- There is currently an issue with the Versioning toolkit itself where it is possible to go down the chain of upgraders with incorrect json: For example, it this case, upgrader 3.3 will upgrade the namespace then pass back to serialisation. That one will fail, send the new modified json for versioning again but 3.0 will not upgrade successfully because the namespace is the new one. There is an argument for always starting from the oldest upgrader and going up instead of doing this two way system.

- For now I therefore decided to rely on adding the versioning to the 3.3 dictionary. Fun fact: 'AllMethodList()' is not loading dll from other UIs to avoid incompatibilities. Given that the Versioning file generator is its own UI and doesn't depend on BHoM_UI, it never load the methods from the BHoM_UI dll.

- I guess that one could also be fixed by adding an exception on BHoM_UI only. For now, I resorted to manually add the methods versioning to the versioning file 

- The versioning for Execute was missing so I ended up writing that one by end (sooo much fun)

- Finally, there was a typo in code of Versioning toolkit where type conversions `ToOld` were actually stored in `ToNew`. I have raised the PR to fix that here: https://github.com/BHoM/Versioning_Toolkit/pull/93. This one can be merged as soon as it is approved as it is a pretty obvious bug fix that would benefit the master as well.

IsakNaslundBh:
Small bug found on the search menu on the `CreateType` component:

image


Everything is functioning as it should, but the property assignment create method is being shown instead of the type in the search list. The global search is working as it should though.

BHoM_UI/Caller/Run.cs
```diff
@@ -0,0 +1,298 @@
...
+                        else
+                            input = InputParams[i].DefaultValue;
+                    }
+                    catch (Exception e)
+                    {
+                        Type originalInputType = InputType(m_OriginalItem, i);
+                        if (originalInputType != null && originalInputType.IsGenericType)
+                        {
+                            UpdateInputGenericType(i);
+                            input = m_CompiledGetters[i](m_DataAccessor, index);
```
IsakNaslundBh, line 139: THis this will resolve the issue

```suggestion
                            input = m_CompiledGetters[i](m_DataAccessor, index);
                            index++;
```

IsakNaslundBh (changes requested):
Found an issue with generic methods with multiple arguments failing to run the first time. After re-running the script, all runs fine.

Think the suggestion below resolves the problem, but will leave it for you to double check @adecler .

Have saved a testscript named "Generics first time run issue.gh" in folder to this PR.

adecler:
@IsakNaslundBh , both bugs should now be fixed. Let me know if you are still having problems.

IsakNaslundBh:
> @IsakNaslundBh , both bugs should now be fixed. Let me know if you are still having problems.

Issues found yesterday all look to be resolved 👍 😄

adecler:
I have removed the `do-not-merge` label on all the UI refactoring PRs as discussed. It is now in a good spot to be tested more heavily by the alpha testers before the creation of the beta installer in 3 weeks.

al-fisher:
Hey @IsakNaslundBh - looking to now merge this block of PRs to unlock further alpha testing - are you happy your requested changes have been addressed? 😄

IsakNaslundBh:
Found an interesting bug where a list input component stopped working after once having been provided a wrong input. Have saved script "List component stops working after wrong input has been used.gh" in refactoring folder.

Issue:
- Component runs fine when only correct inputs are plugged in
- After pluging in incorrect inputs, it throws an error stating that those cant be casted (this is correct)
- Plugging the original, correct, inputs back in to the component it still throws an error, complaining about trying to cast the list, which it can not do. (this is the issue)
- Copy-pasting the component makes it run again

See gif:

ListInputFail

IsakNaslundBh:
> Hey @IsakNaslundBh - looking to now merge this block of PRs to unlock further alpha testing - are you happy your requested changes have been addressed? 😄

And did not spot this before my latest comment. Think the above would be good to get fixed before merging!

al-fisher:
> Think the above would be good to get fixed before merging!

Perfect - thanks!

adecler:
@IsakNaslundBh , it should now be fixed.

This is a weird one: The method itself is not generic so casting the input list is never the right thing to do (as there is no flexibility on the method arguent types anyways). In this case, the incorrect List was forcing a casting on the input type that was still valid when providing the list of beam again. The problem is that the method could not adapt to that new type of list since it was still expecting a list of `IFramingElement`. So, now, I actually check that method is generic before casting the input. There is an argument for making sure that the generic type(s) of the input is/are indeed matching one generic type from the method itself but there are a lot of cases to check here to avoid getting it wrong so I kept it simple for now. 

I think this bug must have been on the master too since I pretty much didn't touch the code around generics but good to have it fixed non-the-less.

IsakNaslundBh (approved):
All issues found so far fixed. Have been on the branch for two weeks now, and things seem to be working fine, especially as the niggily bits have been resolved.

PR #302, opened 2020-09-15
Fix list of possible items in CreateRequest component

### LINKED ISSUE #301: BHoM_UI: CreateRequest exposes all adapter Create methods instead of Request only
#### Description:

As in the description, CreateRequest seems to expose all `Create` methods including the ones that do not output `IRequests`:

image

adecler:

Closes #301

### Test files
Just make sure that the menu of the CreateRequest component only allows to produce `IRequest`

### DIFF
--- a/UI_Engine/Query/Items.cs
+++ b/UI_Engine/Query/Items.cs
@@ -107,7 +107,7 @@ namespace BH.Engine.UI
         {
             return Engine.Reflection.Query.BHoMTypeList()
                 .Where(x => x != null && x.Namespace.StartsWith("BH.oM.Adapters") && !x.IsNotImplemented() && !x.IsDeprecated() && !x.IsEnum && !x.IsAbstract)
-                .Where(x => x.GetConstructors().Where(c => c.GetParameters().Count() > 0).Count() == 0);
+                .Where(x => typeof(BH.oM.Data.Requests.IRequest).IsAssignableFrom(x) && x.GetConstructors().Where(c => c.GetParameters().Count() > 0).Count() == 0);
         }
 
         /***************************************************/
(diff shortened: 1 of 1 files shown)


pawelbaran (dismissed):
LGTM

UI_Engine/Query/Items.cs
```diff
@@ -107,7 +107,7 @@ public static IEnumerable<Type> ConstructableRequestItems()
         {
             return Engine.Reflection.Query.BHoMTypeList()
                 .Where(x => x != null && x.Namespace.StartsWith("BH.oM.Adapters") && !x.IsNotImplemented() && !x.IsDeprecated() && !x.IsEnum && !x.IsAbstract)
```
IsakNaslundBh, line 109: Do not think we should/need to filter only items in the adapter namespace. The `FitlerRequest` for example sits in `BH.oM.Data` and there are other requests available in Structures for example.

```suggestion
                .Where(x => x != null  && !x.IsNotImplemented() && !x.IsDeprecated() && !x.IsEnum && !x.IsAbstract)
```

IsakNaslundBh (changes requested):
The change fixes the issue with non-IRequest showing up, but think the property assignment methods for non-adapterspecific requests are missing, see suggestion below.

UI_Engine/Query/Items.cs
pawelbaran: Hmm, seems to work fine for me functionality wise, even though the code would say something else:

image

IsakNaslundBh: That is only the create methods, not the property assignment methods (automatic creates). That is at least the only ones showing up for me.

pawelbaran: Right. I stopped at the stage where the `Create` methods were needed, so simply added all of them. But yeah, would be great to be able to get rid of them and stick to the auto creates.

IsakNaslundBh (approved):
LGTM

Issue #305, opened 2020-09-16
Grasshopper freeze on startup


IsakNaslundBh:
#### Description:

Not sure if this is a UI issue or a GH issue, but raising here as most code is centralised.

In some occasions, Grasshopper freeze on startup Looking like this:
![image](https://user-images.githubusercontent.com/22005920/93364061-5f64d980-f848-11ea-95ad-9ff2b65b3b4d.png)

Trying to start it again makes GH load up but with limited functionality. For example:
- CreateObejct component not functioning
- Global search menu not functioning

Restarting Rhino and Grasshopper everything again runs with no issue

#### Steps to reproduce:

Will try to keep an eye out for when this is occurring, but has only happened a few times and have still not been able to spot a pattern.

al-fisher:
I just experienced the exact same behaviour. 

- Grasshopper hanging. 
- Rhino responsive and so called `Command: _Grasshopper` again, and Grasshopper opened by same limited BHoM functionality as @IsakNaslundBh notes.
- Close Rhino and restarted - all ran as expected

adecler:
After closing and reopening GH as many times I could handle, I was not able to reproduce the bug. 
Next time, can you hook up Visual Studio and check where the process is hanging ?

adecler:
Looking at this again, I see there is no progress bar under the Grasshopper splash screen.
Was that progress bar present at any time ? Not sure about you but the BHoM is usually one of the last plugin that GH is loading so a bit surprising not to see a progress bar at all. 

If there was never a progress bar, the only thing I can think of is that some of the pre-processing GH does before loading the plugins (and displaying the progress bar) is causing it to hang. Very weird though as I cannot really of something that would cause it work sometimes and sometimes not. Also unclear why it doesn't happen on my machine despite me having probably open and closed GH more than anyone during the UI development.

IsakNaslundBh:
> Looking at this again, I see there is no progress bar under the Grasshopper splash screen.
Was that progress bar present at any time ? Not sure about you but the BHoM is usually one of the last plugin that GH is loading so a bit surprising not to see a progress bar at all.

For me there was a progress bar that filled up and went through all the plugins. This freeze happened after it had filled up. Basically, the progress bar filled up completely, then disappeared, frozen on the screen on the snip.

[...]

PR #306, opened 2020-09-17
Possible fix to Grasshopper sometimes freezing at launch

adecler:

Not sure if it actually solves the problems but this is related to issues #305. 
I suggest we keep that issue open even if we merge this PR though. As I never had that issue on my machine, this PR is just a blind guess at what might be causing the issue. An additional try-catch cannot hurt though 😄 

It is worth mentioning that 
- BHoM_Analytics is running asynchronously so there is not risk there for it to freeze Grasshopper. 
- Usage log files are not accessed during initialisation so no risk on that side either.

### Test files
Keep opening and closing Grasshopper I guess 😋

### DIFF
--- a/BHoM_UI/Global/Initialisation.cs
+++ b/BHoM_UI/Global/Initialisation.cs
@@ -66,12 +66,19 @@ namespace BH.UI.Base.Global
             bool success = true;
             foreach (string file in Directory.GetFiles(@"C:\ProgramData\BHoM\Settings", "*.cfg"))
             {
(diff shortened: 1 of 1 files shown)


al-fisher (approved):
LGTM. Tested and naturally settings file read is still behaving as expected for the case of standard valid .config.

IsakNaslundBh (approved):
Seem fine and sensible to protect here. Agree with all comments in description

PR #319, opened 2020-10-08
Adding null checks to Caller.UpdateInput

### LINKED ISSUE #318: Caller.UpdateInput needs a few extra checks
#### Description:
Dynamo doesn't always provide valid parameters (name and type) to `Caller.UpdateInput` so safer to add a few extra checks.

adecler:

Closes #318

Dynamo doesn't always provide valid parameters (name and type) to `Caller.UpdateInput` so safer to add a few extra checks.
I have also allowed name to be null for additional flexibility on when that method can be triggered.

As soon as this PR is merged, I'll push the changes on the side of Dynamo to the open PR.

#### Test file(s):
A  simple review of the code should be enough here as it is purely adding null checks

### DIFF
--- a/BHoM_UI/Caller/IO_CRUD.cs
+++ b/BHoM_UI/Caller/IO_CRUD.cs
@@ -88,10 +88,10 @@ namespace BH.UI.Base
 
             if (type != null)
             {
-                InputParams[index].DataType = type;
-
                 if (type != InputParams[index].DataType)
                     m_CompiledGetters[index] = Engine.UI.Create.InputAccessor(m_DataAccessor.GetType(), type);
+
+                InputParams[index].DataType = type;
             }
                 
             return true;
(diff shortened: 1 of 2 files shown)


IsakNaslundBh (changes requested):
minor tweaks

BHoM_UI/Caller/IO_CRUD.cs
```diff
@@ -75,20 +75,25 @@ public virtual bool RemoveInput(string name)
...
-                m_CompiledGetters[index] = Engine.UI.Create.InputAccessor(m_DataAccessor.GetType(), type);
-
             if (name != null)
                 InputParams[index].Name = name;
 
             if (type != null)
+            {
                 InputParams[index].DataType = type;
 
+                if (type != InputParams[index].DataType)
```
IsakNaslundBh, line 93: code message was not posted for whatever reason:

This if should always be true given the parameter is assigned just above.

Not sure if the execution order is important, as in, a new compilation should only happen if the input has actually changed, then I would move this above line 90. If not, I think we should jsut delete the if.

adecler: Well, that happened :disappointed:

Now fixed

IsakNaslundBh (approved):
LGTM now

PR #321, opened 2020-10-10
Fix ParamSelector menu getting out of sync in Dynamo

### LINKED ISSUE #320: ParamSelectorMenu only stays in sync if recreated each time
#### Description:
In Grasshopper, the menu is recreated every time so all is good but it is only created once in Dynamo. The problem is that the input params are removed from the component `-` button, the menu doesn't update correctly. 

#### Steps to reproduce:
In Dynamo, click on the `-` button of a CreateObject component (e.g. Point) and then check the input selection menu.

adecler:

Closes #320

See issue

### Test files
Just check the menu after deleting some of the inputs using the `-` button both in Grasshopper and Dynamo.

### DIFF
--- a/BHoM_UI/Menus/ParamSelectorMenu.cs
+++ b/BHoM_UI/Menus/ParamSelectorMenu.cs
@@ -68,8 +68,12 @@ namespace BH.UI.Base.Menus
                     e.Cancel = true;
             };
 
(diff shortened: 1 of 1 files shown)


pawelbaran (approved):
An improvement for Dynamo, Grasshopper side seems to be unaffected. LGTM 👍

PR #323, opened 2020-10-14
Leverage attributes to better control auto-create components

### LINKED ISSUE #205: BHoM_UI: Handling of default values for auto-constructor
As we move more and more to the auto-generated property assignment creates, some things have come to mind that these methods right now wont be able to handle:

- Required properties for the object to be valid (Start and End node for a bar for example)
- Kind of related, but warnings for properties that you generally will want to assign for certain workflows to work (names for properties that are to be pushed to a structural package for example)

There are also some cases where we want to pull default values from some Dataset or similar if no property is provided (this ofc already sits under the valid Create method context, but might be a simple property assignment method, that just checks for default values if nothing is provided and raises a note/warning that this has been done).

Some solutions that come to mind, also from some initial discussion with @al-fisher , is to solve this by some attributes that can be added to the oM. For Required properties we could simply add a `Required` attribute that you could put on properties, and then the auto create component could basically just treat these inputs as non-optional.

For warnings, we could have a similar attribute warning, something like `DefaultValueWarning` attribute that could be used as
`[DefaultValueWarning("The name of the section has not been set. Please be aware that you will need to set the name of this object to be able to push it to most structural packages!")]`

For the case of required/default values, it will be trickier when the required proeprty comes from the base BHoMObject, but that could be solved by making all properties on the BHoMObject virtual (think we almost should do this anyway in the spirit of "everything should be hackable") and then simply overriding them and adding the attributes when needed.

Just some general thought. I know more warnings and instructions as to what is required/needed for certain workflows to work has been requested by several users, and just want to make sure the autocreates does not move us one step away from that. Thankful for any comments/other ideas.

@al-fisher @adecler @rwemay @FraserGreenroyd @alelom @epignatelli @JosefTaylor @peterjamesnugent

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/1020
https://github.com/BHoM/BHoM_Engine/pull/2057
   
Closes #205

see the oM PR for more details

The support for converting deleted Create methods into auto-constructors was already existing in the BHoM_UI code but was broken after the refactoring. It was a small enough fix and relevant enough for this PR that  Ideceided to do it here.

Closes https://github.com/BHoM/Versioning_Toolkit/issues/96
### Test files
Just add those attributes to your favorite class and test that it works in the UI

### DIFF
--- a/BHoM_UI/Caller/Run.cs
+++ b/BHoM_UI/Caller/Run.cs
@@ -125,6 +125,13 @@ namespace BH.UI.Base
                         try
                         {
                             input = m_CompiledGetters[i](m_DataAccessor, index);
+
+                            if (input == null || input.ToString() == "")
+                            {
+                                string warning = InputParams[i].DefaultValueWarning;
+                                if (!string.IsNullOrEmpty(warning))
+                                    Engine.Reflection.Compute.RecordNote(warning);
+                            }
                         }
                         catch (Exception e)
                         {
@@ -141,8 +148,13 @@ namespace BH.UI.Base
                         index++;
                     } 
                     else
+                    {
+                        string warning = InputParams[i].DefaultValueWarning;
+                        if (!string.IsNullOrEmpty(warning))
+                            Engine.Reflection.Compute.RecordNote(warning);
                         input = InputParams[i].DefaultValue;
-                    
+                    }
+                        
                     inputs.Add(input);
                 }
             }
--- a/BHoM_UI/Caller/Read.cs
+++ b/BHoM_UI/Caller/Read.cs
@@ -153,7 +153,7 @@ namespace BH.UI.Base
             // Maybe this is an old Create method ?
             object selectedItem = null;
             if (outputParams.Count == 1)
-                selectedItem = OldCreateMethodToType(json);
+                selectedItem = OldCreateMethodToType(json, inputParams, outputParams);
 
             // If the selected Item is not found, we need to make sure that the component keeps its old inputs and outputs
             if (selectedItem == null)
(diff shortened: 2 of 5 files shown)


BHoM_UI/Caller/Run.cs
```diff
@@ -125,6 +125,13 @@ protected virtual List<object> CollectInputs()
                         try
                         {
                             input = m_CompiledGetters[i](m_DataAccessor, index);
+
+                            if (input == null || input.ToString() == "")
```
IsakNaslundBh, line 129: wonder if this instead of checking for null should be checking some equality of default(T) or similar, as this would always get hit for numeric values etc.

IsakNaslundBh: Replying myself here after another thought. My suggestion of the above would mean raising notes if a 0 value has been provided, which for many cases is very valid, so unless we move where the check is made to check if any value has been provided, would rather keep this as is.

IsakNaslundBh (changes requested):
Tested with Bar class (as example):

image

Default value warning comes through as expected, and disappears when a section is plugged in. 👍 

The Required attributes are not behaving as I would expect though. I would Expect the component to go orange if no value is provided to the required inputs. Right now it seems as the difference they make is making it impossible to remove them from the input side:

image


I would expect the behaviour of the parameters without default values in the engine:

image

Basically, without any nodes, the bar is not defined. The nodes are required.

Also tested the `NoAutoConstructor` on BarUniformlyDistributedLoad and it succefully hid the auto create 👍

BHoM_UI/Caller/Run.cs
adecler: @IsakNaslundBh . Yes the reasoning was exactly around the `0` value. We always add a extra `OR` check later if we missed something

adecler:
@IsakNaslundBh , good catch. 
This is actually a Grasshopper issue.  
I have raised the PR here: https://github.com/BHoM/Grasshopper_Toolkit/pull/556

IsakNaslundBh (approved):
>@IsakNaslundBh , good catch.
This is actually a Grasshopper issue.
I have raised the PR here: BHoM/Grasshopper_Toolkit#556

Perfect! This now does what I expect it to in GH. Ofc good if anyone else want to have a review as well! Also, good if someone can test the other UIs @kayleighhoude @pawelbaran

adecler:
> Also, good if someone can test the other UIs @kayleighhoude @pawelbaran

Ha, I can answer that one: works in Dynamo, doesn't in Excel. Just that Excel doesn't really have much support for removing inputs right now so will have to put something in place (out of the scope of this sadly).

PR #325, opened 2020-10-15
Make sure success output is false when active is false on adater components

### LINKED ISSUE #324: Push success refresh
#### Description:
Currently **Push** remembers previous state even after turning off the **active** toggle. This is helpful to further process the resultant objects that were being pushed, but the **success** output is misleading and can't be used to trigger execution of the following **Push** components.

MicrosoftTeams-image (9)

#### Steps to reproduce:
Turn on the adapter and **active** toggle and then turn it off - the output **success** will be showing **True**.

#### Expected behaviour:
In order to be able to use the **Success** field as future action trigger, I would prefer it to turn to **False** after **active** is changed to **False**. Otherwise it can lead to errors visible on the screen below. **Objects** should remain as it is.

MicrosoftTeams-image (10)

Message: 
_Error: Object reference not set to an instance of an object. ... called from BH.UI.Base.Components.PushCaller.Push(BHoMAdapter adapter, IEnumerable`1 objects, String tag, PushType pushType, ActionConfig actionConfig, Boolean active)_

adecler:

Closes #324

See issue

### DIFF
--- a/BHoM_UI/Caller/Run.cs
+++ b/BHoM_UI/Caller/Run.cs
@@ -52,7 +52,7 @@ namespace BH.UI.Base
 
             // Execute the method
             object result = m_CachedResult;
-            if (m_CachedResult == null || ShouldCalculateNewResult(inputs))
+            if (m_CachedResult == null || ShouldCalculateNewResult(inputs, ref result))
             {
                 try
                 {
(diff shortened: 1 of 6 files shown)


alelom (approved):
LGTM! Thanks for this one.

PR #326, opened 2020-10-16
Align to change in SetProperty method name

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_Engine/pull/2066

   
See https://github.com/BHoM/BHoM_Engine/pull/2066 for details

### Test files
Just make sure that the `SetProperty` component is still working for both new and saved ones. Note that I have also tested for setting of subproperies and it still works fine (e.g. `Location.X`). The other cool thing is that the `SetProperty` component is not limited to BHoM objects anymore (before, you couldn't set a property of a BHoM Point for example). That aligns better with previous changes on things like `ToJson`

### DIFF
--- a/BHoM_UI/Components/Engine/SetProperty.cs
+++ b/BHoM_UI/Components/Engine/SetProperty.cs
@@ -51,7 +51,7 @@ namespace BH.UI.Base.Components
         /**** Constructors                ****/
         /*************************************/
 
-        public SetPropertyCaller() : base(typeof(Engine.Reflection.Modify).GetMethod("PropertyValue")) { }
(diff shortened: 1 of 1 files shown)


alelom:
Although for object that have a valid `Set` for their property this works nicely, attempting to use `SetProperty` for a property without a setter crashes Rhino for me:
image

Script saved here:
https://burohappold.sharepoint.com/:f:/s/BHoM/Esemk8ahTL1Lr3y4XAPZi8IBv6p_zx2r03oKFLvRTuybdw?e=qagAvB

adecler:
Thanks @alelom , this should now be fixed (commit in the Engine)

alelom (approved):
Now it's perfect 🚀 
image

PR #331, opened 2020-11-04
Adding GetEvents component

### LINKED ISSUE #235: Enable a generalised way to access any warnings and/or error on BHoM load in any UI
From workshop on Excel - idea was mooted to enable access via Ribbon to any logged warnings - shown to user as an "exclamation mark" style feedback.

To avoid using a pop up with is more intrusive 

@adecler suggested there is already some form of access via a "log" method in Reflection Engine. 

To be formalised and then given access from each UI as appropriate

@adecler @awakeman

adecler:

Closes #235

Adding component that better expose the ability to access the error, warning, and note events that have occurred until now. 
At the moment, the only (optional) input I have for it is the timestamp after which we want the events. Not sure that it is clear so here's an example:

image

Happy to add more inputs if you feel necessary.

Ass agreed, it's super simple for now so no advance popup window or anything of the sort.

You will need the PR on each UI to test it there.

### UI PRs
- Excel: https://github.com/BHoM/Excel_Toolkit/pull/264
- Dynamo: https://github.com/BHoM/Dynamo_Toolkit/pull/274
- Grasshopper: will be pushed as soon as the current PR on GH is merged.


BHoM_UI/Components/Engine/GetEvents.cs
```diff
@@ -0,0 +1,87 @@
...
+        /*************************************/
+
+        public GetEventsCaller() : base(typeof(GetEventsCaller).GetMethod("GetEvents")) { }
+
+
+        /*************************************/
+        /**** Public Method               ****/
+        /*************************************/
+
+        [Description("Get all events (errors, warnings, and notes) occuring on BHoM components.")]
```
pawelbaran, line 66: `since` input is missing

pawelbaran (commented):
The functionality works fine, as mentioned in https://github.com/BHoM/Excel_Toolkit/pull/264 and https://github.com/BHoM/Dynamo_Toolkit/pull/274, only one description is missing in this PR (see the code comment).

I've noticed the events contain startup messages, incl. _cannot load blabla_ stuff that I guess will not be too useful for the regular user and can raise questions - I know these can be hidden by specifying `since` input, but wouldn't it be worth hiding them altogether (general question, I do not have a strong opinion on that)?

FraserGreenroyd (dismissed):
Code LGTM, I would agree with @pawelbaran that a `since` input attribute would be nice but I'm also happy without it at this moment.

In terms to the other point @pawelbaran raised, I would agree it could be confusing to see errors from start up when they evidently were ok to continue running, but as a first iteration of the component I'm happy with it. I dunno how easy it is to have an event listening to when the UI is fully loaded and then default the `since` input to be after that time? I'd imagine it's not terribly easy which is why, as I say, I'm happy with its current functionality rather than holding this up for a minor UX improvement :smile: (well, minor in my mind 😝 )

adecler:
Thanks both.

I have added the description for the input.

I think it is really important to have the startup events easily available as they were the whole reason why we even started the conversation on an central error listing solution.  I agree it would better to have them listed separately though. So a fourth output for start up events would be nice to have. 

I only have startup messages in Excel but this is now looking like this:
image

pawelbaran (approved):
LGTM, separating the startup events is great!

PR #333, opened 2020-11-06
Code updated to the changes introduced in BHoM #1068 (references updated)

pawelbaran:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1068
https://github.com/BHoM/BHoM_Engine/pull/2132
   
Closes https://github.com/BHoM/BHoM/issues/1067

### Test files

See https://github.com/BHoM/BHoM_Engine/pull/2132.

### Changelog

- code updated to the changes made in https://github.com/BHoM/BHoM/pull/1068

### Additional comments

Both this PR and #331 modify the same file, but GitHub should be able to deal with rebasing both (at least I believe so 🙊). However, if you are afraid that will not be the case @adecler, then let's merge #331 first, regardless of which one is ready sooner.

### DIFF
--- a/BHoM_UI/BHoM_UI.csproj
+++ b/BHoM_UI/BHoM_UI.csproj
@@ -73,6 +73,10 @@
     <Reference Include="PresentationFramework">
       <Private>False</Private>
     </Reference>
+    <Reference Include="Quantities_oM">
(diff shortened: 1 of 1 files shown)


adecler (approved):
LGTM

PR #334, opened 2020-11-12
Make it easy to clear the value of a property

### LINKED ISSUE #141: BHoM_UI: setting a empty list from a C# component in SetProperty
#### Description:
To remove openings from a panel I'm setting a empty Openings list `List<Openings>` with `SetProperty` in Grasshopper. This used to work but now i first throws the error: `GetDataList() can only be called on a parameter with access set to GH_ParamAccess.list`

Thus I change it to List access, where it throws this error: `
Unable to cast object of type 'System.Collections.Generic.List`1[BH.oM.Structure.Elements.Opening]' to type 'BH.oM.Structure.Elements.Opening'.`

Then I change it back to Item access where it works...
and then I exit my cluster and it's broken again

I tried this outside of an cluster as well and it broke upon reopening the document.

It can also be "fixed" by re-attaching any of the Input wires, but updating the data in the objects did not.

Duplicating the Component re-broke it again

#### Test file(s):
[TestFile](https://burohappold.sharepoint.com/:u:/s/BHoM/EVQVAm4UWcROi0HRY9YQgmkBxaSxVLqXoeM70t_K4rPQyA?e=aGOs9e)

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_Engine/pull/2142

   
Closes #141 

With this PR, it is now easy to clear the value of a property by leaving the `value` input empty. Here's how it behaves:
- Value types like primitives and enums are set to their default value (value types cannot be null)
- Any IEnumerable is set to an empty version of that type.
- Strings are set to an empty string (it could be set to null but I think this is the only case where it makes more sense not to)
- All other objects are set to null.

Looks like a lot of cases listed like that but I found it pretty intuitive when actually used. Let me know if you disagree.

### Test files
https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_UI/BHoM_UI-Issue141-SetPropertyDefault.gh?csf=1&web=1&e=h7RH22

### Additional comments
Note that the issue @kThorsager was complaining about related to the list type disappearing when copying the component has been solved during the refactoring of the UI.

### DIFF
--- a/BHoM_UI/Components/Engine/SetProperty.cs
+++ b/BHoM_UI/Components/Engine/SetProperty.cs
@@ -80,7 +80,12 @@ namespace BH.UI.Base.Components
                         {
                             PropertyInfo propInfo = objType.GetProperty(propName);
                             if (propInfo != null)
-                                m_CompiledGetters[2] = Engine.UI.Create.InputAccessor(m_DataAccessor.GetType(), propInfo.PropertyType);
+                            {
+                                Type propType = propInfo.PropertyType;
+                                if (propType.IsValueType)
+                                    propType = typeof(object);
+                                m_CompiledGetters[2] = Engine.UI.Create.InputAccessor(m_DataAccessor.GetType(), propType);
+                            }    
                         }
                     }
 


pawelbaran (changes requested):
I went through this PR as well as https://github.com/BHoM/BHoM_Engine/pull/2142 - please consider this review as combined for both.

I see two points I am not sure about. Firstly, the enums get reset to their first value (value equal to 0 I guess), which might be confusing as sometimes the desired default might not sit under 0. Not sure if it would not be possible, but maybe the properties could be reset not to zero/null, but to their default values instead (and zero/null if there is no default value)?

Second issue is purely technical, I get an error after opening the file:
image

adecler:
>I see two points I am not sure about. Firstly, the enums get reset to their first value (value equal to 0 I guess), which might be confusing as sometimes the desired default might not sit under 0. Not sure if it would not be possible, but maybe the properties could be reset not to zero/null, but to their default values instead (and zero/null if there is no default value)?

That would mean a different behaviour for enums than for all other types as this always sets to the equivalent of null, not to the initial value of the property. Keep in mind that this is purely to provide a behaviour in every case that is consistent. If we want to reset to the initial value of the property, we have to do it for every type. It then requires knowledge of the what those values are as they are not always null. It also means that, if the initial value is not null, there would actually be no way of setting it to null even with this component. I also don't expect the user to leave the `value` input empty when it makes no sense to do so (and if they do so, I have a ready made answer for them 😉 ).

adecler:
> Second issue is purely technical, I get an error after opening the file:

I forgot to save it with the `value` input saved as a list. Changing that, saving, and reopening the file works perfectly fine for me. Can you confirm it is the case for you too ?

pawelbaran (approved):
Following today's discussion happy with the changes. However, I just thought of a neat improvement: would it be possible to set the property value to null regardless its type (item/list/tree), so that the user would not need to change input type through the UI (the error mentioned by me [here](https://github.com/BHoM/BHoM_UI/pull/334#pullrequestreview-529039523) could be avoided)?

PR #336, opened 2020-11-17
BHoM_UI: Detect startup using UTC time instead of local time

### LINKED ISSUE #335: Event logging gone mad
#### Description:

That is a weird issue that I've found when testing https://github.com/BHoM/Revit_Toolkit/pull/904 (happens only on this branch) with the following files from [here](https://burohappold.sharepoint.com/sites/BHoM/Installers/Forms/AllItems.aspx?RootFolder=%2fsites%2fBHoM%2fInstallers%2fScripts%2fBuroHappold%5fBHoM%5fv3%2e3%2f0001%5fRevit%5fToolkit%2fPull&FolderCTID=0x012000181C071E8B6D1E438B59C87DA36B9302):
- BuroHappold_BHoM_v3.3.beta_PullFramingProfiles.rvt
- BuroHappold_BHoM_v3.3.beta_PullFramingProfiles.gh

1. After loading the .gh script, `GetEvents` does not find startup issues (says there was none):
image
2. After activating the adapter in Revit (in _BHoM_ tab => _Activate_), I have selected the element with Id 1454194 (_Manage_ => _Select by ID_) and ran `Pull` in the GH script. A single warning pops up in the pull component (correct), but if you use `GetEvents`, one can see it got logged in startup warnings!
image
image
3. After selecting an element with Id 1394404 in Revit and reactivating the `Pull` in .gh script, the startup warnings show up in the `Pull` component!
image
At the same time, `GetEvents` finally catches all startup warnings:
image
4. After pulling same selection again, the pull goes fine without any warning (which is the expected behaviour).

#### Summary:
The startup warnings originally do not show up in `GetEvents` (that can be easily tested by simply opening an empty script and checking `GetEvents`), but they somehow get caught once certain methods get triggered (in case of my Revit_Toolkit branch, it is `BH.Engine.Spatial.Create.FreeFormProfile` from within a convert) plus at the same time they get logged by an innocent component (in my case `Pull`). What is more, my `Pull` events get logged by `GetEvents` as startup ones. **_That is weird_**.

Now I only speculate, but this could have something to do with Revit being a separate process, or my time zone being somehow messed up. No clue, I can only speak for what I see 😉

Looping @IsakNaslundBh @alelom @al-fisher in as not sure how wide will this issue spread plus I think it may be quite critical.

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/1092

   
Closes #335

It seems that some events recorded in Revit are using UTC time instead of local time despite using `DateTime.Now` and not `DateTime.UtcNow`. So detecting startup events can only be done reliably using UTC time everywhere. 

It is important to note, that I keep using local time for the `since` input of `GetEvents` because this is something provided by the user.

### DIFF
--- a/BHoM_UI/Components/Engine/GetEvents.cs
+++ b/BHoM_UI/Components/Engine/GetEvents.cs
@@ -77,8 +77,8 @@ namespace BH.UI.Base.Components
             List<Event> startupEvents = new List<Event>();
             if (Initialisation.CompletionTime != null)
             {
-                startupEvents = events.Where(x => x.Time <= Initialisation.CompletionTime).ToList();
-                events = events.Where(x => x.Time > Initialisation.CompletionTime).ToList();
+                startupEvents = events.Where(x => x.UtcTime <= Initialisation.CompletionTime).ToList();
+                events = events.Where(x => x.UtcTime > Initialisation.CompletionTime).ToList();
             }
 
             if (since != null)
--- a/BHoM_UI/Global/Initialisation.cs
+++ b/BHoM_UI/Global/Initialisation.cs
@@ -57,7 +57,7 @@ namespace BH.UI.Base.Global
 
             success &= LoadToolkitSettings();
 
-            CompletionTime = DateTime.Now;
+            CompletionTime = DateTime.UtcNow;
 
             return success;
         }
--- a/BHoM_UI/Global/SearchMenu.cs
+++ b/BHoM_UI/Global/SearchMenu.cs
@@ -62,7 +62,7 @@ namespace BH.UI.Base.Global
         public SearchMenu()
         {
             PossibleItems = GetAllPossibleItems();
-            Initialisation.CompletionTime = DateTime.Now;
+            Initialisation.CompletionTime = DateTime.UtcNow;
         }
 
         /*************************************/


FraserGreenroyd (approved):
Code LGTM in line with the oM addition

FraserGreenroyd:
@BHoMBot check installer

pawelbaran (changes requested):
The original issue still persists. Even worse, in point 3. of it (pulling of the element 1394404), the startup warnings finally appear, but not in the startup section:

image

@FraserGreenroyd could you please try opening an empty GH doc on this branch and check if `GetEvents` logs startup messages under startup output? That is the 1st thing that already goes wrong on my build (see point 1. in https://github.com/BHoM/BHoM_UI/issues/335).

FraserGreenroyd:
> @FraserGreenroyd could you please try opening an empty GH doc on this branch and check if `GetEvents` logs startup messages under startup output? That is the 1st thing that already goes wrong on my build (see point 1. in #335).

Interesting, I'll take another look 😕

adecler:
@pawelbaran , everything actually works exactly as intended.

The issue is that the revit toolkit is not front-loading the BHoM methods so those will only be extracted the first time that method list is needed. In this case, when a call to `RunExtensionMethod` is made:

image

If you want to test that for yourself, keep in mind you need to attach VS to Revit and not Grasshopper.

pawelbaran:
@adecler do you mean there is zero startup messages in Grasshopper itself? Opening GH alone, startup warnings are empty:

image

Coming back to Revit, I have just confirmed your findings, looks like the problem lives inside `RevitListener` - I have transferred the issue to Revit_Toolkit and will resolve it there (I will simply suppress the startup warnings, I am 99.99% sure they will never be relevant by any means).

PR #352, opened 2021-01-07
BHoM_UI: Prototype for the new RunCode component

### LINKED ISSUE #137: BHoM_UI: Improve responsiveness and navigability of the menus/dropdown
#### Description:

Some context menu already start taking too long to open (e.g. Query component). Same thing is true for The Data drop-down component (e.g. RHS sections for the UK_SteelSections). 

When those menus are too tall for the screen, it is also quite slow to scroll up and down (at least in Grasshopper). We need to make sure this is handled better

A few things need to be implemented to improve user experience:
 - progressive loading of the menus (e.g. only two levels down the current one)
 - scroll bar available once the menus are too tall
 - menus loading quickly even when tall (especially true for the drop-down in GH but might be true elsewhere)
 - Add a search item in the context menus for drop-downs

### LINKED ISSUE #188: BHoM_UI: Ability to open the code from any function in The Engine
### This issue has been archived 

Historic issue archive can be found [here](https://burohappold.sharepoint.com/sites/BHoM/01_Master%20File/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F01_Master%20File%2F999_BHoM_GitHub_History_Archive)

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_Engine/pull/2257
https://github.com/BHoM/CSharp_Toolkit/pull/40
   
Closes #188

This is an initial prototype for the RunCode component. Emphasis on the word 'prototype'. This is basically the code I created during the Christmas break to test the viability of the idea. A lot will change but this is nice to get this out of my computer and into your hands for initial testing. 

Closes #137 

Since the menu lists ALL the methods in the BHoM it was opening very very slowly in Grasshopper. So I took the opportunity to also address an old issue from 2019 😄 . With this, the menu are created progressively and are therefore a lot more responsive.

### Test files
This is all new stuff and in a prototype phase so don't worry too much about this one. As long as it works on simple cases I am happy. Personally, this is a component that I intend to use straight away so pretty keen to get that in the master even as a prototype instead of having a WIP PR that stays there for weeks.

**Edit:** Just make sure you also recompile Grasshopper and/or Dynamo PR as well

A few things you can expect to be able to do though:
- See the source code of BHoM methods
- Modify that code and recompile should modify the behaviour of the component itself (including name, inputs, and outputs)
- If the code doesn't compile, the compilations errors will be displayed in the bottom panel.
- You can see the list of references used by that code on the second tab in the bottom.
- If the code changes you make require new references, you can add them by clicking on `+ Add reference...` in the reference tab. Just don't forget to add the `using` in the code 😄 .
- If you open the code editor before selecting a BHoM method, it will create a template for you to play with.
- Code that uses private methods will add them as well
- This is a new code generated on top of the loaded BHoM assemblies, it doesn't override it. So sometimes, I had to modify to declaring type so I can still reference the original one when calling other methods.

### Changelog
- Prototype for the new RunCode component
- Improved responsiveness of menus


adecler:
@FraserGreenroyd , it looks like AppVeyor cannot find `CSharp_oM` and `CSharp_Engine` but it looks like it doesn't even try to compile them. Unless, of course, I am blind 😄

FraserGreenroyd:
@BHoMBot check core

rolyhudson:
Very nice. Seems like this came up just days ago in a conversation that started with a proposal of adding a BHoM scripting  editor, so impressive to see this here already. 👍 

Had a good look at this today  - looking at it from a point-of-view of wanting to refactor a specific method from within the UI.
Very happy to keep discussing and testing this out.

Some observations and thoughts:

- I noticed that the first click on _Select an item_ gives you a menu primarily to the `Create` methods. 
- I could edit and re-run a `Create` method as I was expecting. 
- I could not save my changes. 
- I'm not sure of the implementation complexity but unless I misunderstood the intent I think the ability to save should be essential (working using this on multi-sessions, sharing a proposed refactor with others via a script file).
- To find the `Modify` method I was interested in I had to dig a little - within that first menu list there is an item _NonBHoMObjects_ and this seems to give access to the `Query`, `Modify` and other Engine classes.
- Maybe there a menu structure + hierarchy discussion?
- Once in the editor with a `Query` or `Modify` the class shows up with a prefix underscore like `public static partial class _Query`
- I could not Rebuild see image below
Screenshot 2021-01-07 173414

adecler:
> - I noticed that the first click on _Select an item_ gives you a menu primarily to the `Create` methods. 
> - To find the `Modify` method I was interested in I had to dig a little - within that first menu list there is an item _NonBHoMObjects_ and this seems to give access to the `Query`, `Modify` and other Engine classes.
> - Maybe there is a menu structure + hierarchy discussion or access via a search box?

This is an issue with the menus themselves (unrelated UI code to this). I noticed a few others things like generics name are spread over multiple levels (See Tree<T> for example). So something to sort out on a separate PR.



> - I could not save my changes. 
> - I'm not sure of the implementation complexity but unless I misunderstood the intent I think the ability to save is essential to its usage (working using this over multi-sessions, sharing a proposed refactor with others via a script file...).

It is essential. But remember this is a prototype, not a finished product. This PR is merely to get the experiment I did on Christmas out of the way until I come back to it later this quarter. I didn't do any saving system because the code is going to be centralised as a VS project on the script level so I there was no point spending time implementing a temporary saving per component that would have to be replaced soon.

> - Once in the editor with a `Query` or `Modify` the class shows up with a underscore prefix: `public static partial class _Query`

Yep, that's normal. The code is modified so that it can be compiled. The `BH.Engine.Audience.Query` type already exists in the BHoM. If you want to call some of its other methods you cannot obfuscate it by creating a type with exactly the same name and namespace. In the long run, the editor will highlight the parts of the code that have been modified to make the code compile. Another example is that addition of the type name in front of the calls to method of the replaced type (since the type we create here is considered as a completely different type by the compiler, we have to specify the type that method belongs to).

> - I could not Rebuild on the `Query` or `Modify` methods I looked at, see image below

It compiles if you delete the `Query.` on line 17. This is another example of the need to tweak the code to make it compile. In this case, the addition of the type name in front of the calls to method of the replaced type (since the type we create here is considered as a completely different type by the compiler, we have to specify the type that method belongs to). For now I just do a simple string replace and it just so happens that the method is calling a method with the same name so it gets confused and adds `Query.` in front if its method name. This will go away when I use Roselyn as I will manipulate the code directly on the syntax tree instead. For now, I just went cheap and easy so works for most cases but not all.

adecler:
Quick reminder that this PR is just a temporary prototype, not a finished product. As long as it doesn't break anything, I am happy to merge. It will be properly worked on in a few sprints once the other critical tasks are out of the way. In the meantime, I didn't want this code to sit on my computer or in a PR that will not merge anymore in a few weeks.

alelom (approved):
Really cool prototype! 😄

rolyhudson (approved):
I am approving this prototype following my experiments with this and discovery of no breaking changes.

adecler:
@BHoMBot check installer
@BHoMBot check copyright

adecler:
@BHoMBot check installer

adecler:
@BHoMBot check installer

adecler:
@BHoMBot check copyright

adecler:
@BHoMBot check copyright

FraserGreenroyd:
@BHoMBot check copyright compliance

PR #361, opened 2021-01-23
Handling of Modify methods returning void

### LINKED ISSUE #360: Provide support for Modify methods returning void
#### Description:
With the changes agreed in https://github.com/BHoM/admin/issues/11, components corresponding to a Modify method returning void should provide an output parameter for the modified object.

adecler:

Closes #360

See issue.

### Test files
Not many Modify methods returning `void` right now but I tested it successfully on this:

image

I suggest you either test it on some "hand-made` Modify methods or some that you are familiar with.

### DIFF
--- a/BHoM_UI/Components/Engine/Modify.cs
+++ b/BHoM_UI/Components/Engine/Modify.cs
@@ -27,6 +27,8 @@ using System.Text;
 using System.Threading.Tasks;
 using System.Reflection;
 using BH.Engine.Base;
+using BH.Engine.Reflection;
+using BH.oM.UI;
 
 namespace BH.UI.Base.Components
 {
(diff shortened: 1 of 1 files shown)


BHoM_UI/Components/Engine/Modify.cs
```diff
@@ -63,15 +65,52 @@ public ModifyCaller() : base()
...
+                if (firstParam != null)
+                {
+                    Type returnType = firstParam.ParameterType;
+                    OutputParams = new List<oM.UI.ParamInfo> { new ParamInfo {
+                        Name = "result",
+                        DataType = firstParam.ParameterType,
+                        Description = "Modified copy of the input object.",
+                        Kind = ParamKind.Output,
+                        IsRequired = true
+                    }};
```
FraserGreenroyd, line 100: This seems to be overriding provided `[Output]` attributes when the method is void.

I did a local change to this, and this provided the information back from the `[Output]` attribute. However, it did then put in the message "This is a Void" - I'm guessing that needs an update in the Reflection_Engine to handle that?

That said, I recognise you probably did it this way for a reason @adecler , so just a suggested change in case this is gonna come in with an Engine update :smile:

```suggestion
                    Type returnType = firstParam.ParameterType;
                    string name = method.OutputName();
                    OutputParams = new List<ParamInfo> {
                        new ParamInfo
                        {
                            Name = (name == "") ? returnType.Name.Substring(0, 1) : name,
                            DataType = firstParam.ParameterType,
                            Description = method.OutputDescription(),
                            Kind = ParamKind.Output,
                            IsRequired = true
                        }
                    };
```

FraserGreenroyd (changes requested):
I ran this through with the following scenarios:

 - A method from Environment Engine returning void, checking upstream data isn't changed and output is correct
 - A method from Environment Engine not returning void

Both scenarios worked as expected with this change. My biggest concern is the loss of `output` documentation from this change.

I've made a suggested change based on what I could find from `base.SetOutputs()` and what seemed to work on my machine, but I have little UI knowledge so this might not be the right approach. I also recognise that making the suggested change turns the output to look like this:

image

Where it primarily says `This is a Void` isn't entirely accurate from a user perspective (though 100% accurate from a code perspective :smile: ), though as mentioned I suspect that's a fix needed in Reflection_Engine as it forms part of the `OutputDescription()`.

Happy for an issue to be raised to fix the descriptions later - and if so I'm happy to convert this to an approval review as functionally everything worked as I would expect and we can handle the descriptions later (before beta though please :smile: ). Will leave to you @adecler for best judgement on how/when you want to handle that :smile:

BHoM_UI/Components/Engine/Modify.cs
adecler: To be honest, my logic was that all out put description would be very similar and basically a rewording of the method description. So I opted for a standard simple description to apply all the time. As we said before, having a description attribute for the sake of having one even if it adds nothing is not exactly a great feature 😄

FraserGreenroyd: Ok in that case I'm happy to change this to an approval review :smile: I presume we'll still be changing the `This is a Void` in reflection though? 👼

FraserGreenroyd (approved):
See discussion on output description, functionality working well :smile:

FraserGreenroyd:
@BHoMBot check installer

FraserGreenroyd:
@BHoMBot check copyright compliance

BHoM_UI/Components/Engine/Modify.cs
adecler: Can do. I have raised a separate issue in the engine: https://github.com/BHoM/BHoM_Engine/issues/2287

FraserGreenroyd: Perfect :smile: do you want me to merge this when the checks are complete (assuming they don't finish before you head off) or want to wait for extra reviews?

adecler: I ma happy for whoever to merge once all the checks have passed

adecler:
@BHoMBot check copyright

PR #371, opened 2021-02-17
Fix SetProperty for list nested properties

### LINKED ISSUE #365: SetProperty method struggles with nested properties
#### Description:

image
SetProperty refuses to set the `Reinforcement` via `SectionProperty` with "GetData() can only be called on a parameter with access set to GH_ParamAccess.item" error. It works well with the `MinimumCover` property or with the isolated `SectionProperty`.
(I've changed the "value" input to "List access"). What is more it worked fine few hours ago.

#### Steps to reproduce:

See the test file.

#### Expected behaviour:

Should set the `Reinforcement` in the `SectionProperty`.

#### Test file(s):

[SharePoint](https://burohappold.sharepoint.com/:f:/s/BHoM/EuleqCwV7x1Gl_Rr0G9o2F4B526jGj-nMrGDdw0ihZXvOQ?e=kexUYA)

adecler:

Closes #365

### Test files
[SharePoint](https://burohappold.sharepoint.com/:f:/s/BHoM/EuleqCwV7x1Gl_Rr0G9o2F4B526jGj-nMrGDdw0ihZXvOQ?e=kexUYA)

image

### Additional comments
It is worth noting that something like `bar.SectionProperty.Reinforcement` would not actually compile in C# because `SectionProperty` of a `Bar` is an `ISectionProperty`. That interface doesn't have a `Reinforcement` property. So technically, there is an argument for returning an error when trying to execute the same in Grasshopper. Since there is however a workaround to give the users what they need, I think it is better to do so.

### DIFF
--- a/BHoM_UI/Components/Engine/SetProperty.cs
+++ b/BHoM_UI/Components/Engine/SetProperty.cs
@@ -27,6 +27,7 @@ using System.ComponentModel;
 using System.Reflection;
 using BH.Engine.Base;
 using System.Collections.Generic;
+using BH.Engine.Reflection;
 
 namespace BH.UI.Base.Components
 {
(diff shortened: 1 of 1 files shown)


adecler:
@BHoMBot check copyright

adecler:
@BHoMBot check installer

LMarkowski (approved):
Fair point in the additional comments. I like the proposed solution. I think allowing to set properties this way is just intuitive for a GH user.

PR #376, opened 2021-02-24
Fix Property access type for specific case of accessing Object.List.Property

### LINKED ISSUE #375: Grasshopper data access error introduced with recent change to SetProperty
#### Description:

Apologies I didn't get a chance to test prior to the SetProperty updates here https://github.com/BHoM/BHoM_UI/pull/372, but I've found another problem introduced. Seems this is now enforcing a specified data access type rather than allowing that to be determined by the user/objects. 

200217 Alpha (expected behaviour)
image

200223 Alpha (same workflow with sample objects)
image

#### Test file(s):

[Test File](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/BHoM_UI/%23375-SetPropertyDataAccess/SetMatProp.gh?csf=1&web=1&e=HLtUWZ)

FYI @enarhi

adecler:

Closes #375

Fix the case where the user try to set a property of object inside a list property. Example: setting the property `Layers.Material` of a `Construction`.

This also added a fallback object type for case where the user is trying to set properties in the CustomData for example. 

### Test files
Use the one provided in the issue.

I have also added some additional check where the List property is not following an interface property just to be sure (as those two scenarios are treated slightly differently).

I would recommend also testing anything that comes to mind or existing files as it seems there are still cases I don't think about (as this PR proves). I have, for example, re-tested the file from this PR: https://github.com/BHoM/BHoM_UI/pull/371

### DIFF
--- a/BHoM_UI/Components/Engine/SetProperty.cs
+++ b/BHoM_UI/Components/Engine/SetProperty.cs
@@ -28,6 +28,8 @@ using System.Reflection;
 using BH.Engine.Base;
 using System.Collections.Generic;
 using BH.Engine.Reflection;
+using System.Linq;
+using System.Collections;
 
 namespace BH.UI.Base.Components
 {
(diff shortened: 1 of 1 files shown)


adecler:
@BHoMBot check copyright

adecler:
@BHoMBot check installre

michaelhoehn (approved):
LGTM Thanks for the quick solution here.

PR #394, opened 2021-05-28
Migrate ISettings to the oM.Base namespace

### LINKED ISSUE #393: Move ISettings to the oM.Base namespace
#### Broken rules:
The `ISettings` interface is currently defined in the `oM.UI` namespace but could very much be used as settings for an adapter for example. 

#### Suggestions to restore compliance:
Move the `ISettings` to the `BH.oM.Base` namespace

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/1249

   
Closes #393 

To add a bit of context: Adapters like the BuildingPerformance adapter will need to depend on ISettings to store the link to the database. So it is important that we don't add dependencies between adapter toolklits and the BHoM_UI repo (even if `ISettings` is in `oM.UI`, this is still a repo-level dependency we would like to avoid)

### Test files
Just make sure everything still compiles and that an installer can be created.


adecler:
@BHoMBot check required

FraserGreenroyd (approved):

FraserGreenroyd:

PR #401, opened 2021-08-27
Adding link to source for CreateDataCaller

### LINKED ISSUE #400: BHoM_UI: Add clickable link to source for DataComponents
#### Description:

Many datasets contain a link to the source from which the data has been taken. This link, if available, is stored on the Source object of the serialised dataset.

The link shows up when you hoover over the dataset in the description, but there is no easy way to go to it, besides typing it in yourself.

Would be really useful to be able to have a clickable link on the DataComponent menu that takes you to the source (if available) to more easily determine where the dataset is coming from and evaluate if the information is what you are after.

IsakNaslundBh:

Closes #400 

Checking if the dataset contains a SourceLink, and if so, add a link to it in the menu. Really useful to have quick direct access to the source of the data, when available.

Added the functionality for both the Toolstrip and WPF, but was not able to test the WPF as I do not know where it is being called from. Tried in dynamo, but looked as if the menu is only triggered once, when you add the component, and not again when a dataset has been selected. Happy to remove the WPF implementation if we think that is a better way to go.

### Test files

Any of the UK steelsection or EU material datasets should have working links:

image

### DIFF
--- a/BHoM_UI/Components/oM/CreateData.cs
+++ b/BHoM_UI/Components/oM/CreateData.cs
@@ -34,6 +34,7 @@ using BH.Engine.Data;
 using BH.Engine.Serialiser;
 using System.Windows.Forms;
 using BH.oM.Base;
+using System.Windows;
 
 namespace BH.UI.Base.Components
 {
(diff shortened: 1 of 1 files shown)


adecler (commented):
This is working as expected in Grasshopper.

I would however recommend to make some changes before merging. There is no guarantee that other UIs would behave like GH and recreate the menu every time. You already have an example in that with Dynamo. To fix that, you need to also update the menu when the `FileName` is changed. There are a few places you can override to take care of it: 
- `SetItem` method
- `SetComponentDetails` method
- setter of the `SelectedItem` property. 

The `SetComponentDetails` method is probably the simplest place to do it though.

adecler (changes requested):
A few little things to correct:
- `AddToMenu(ToolStripDropDown menu)` and `AddToMenu(System.Windows.Controls.ContextMenu menu)` are public methods. So they could be called directly without going through `AddToMenu(object menu)` first. So you need to override them. Otherwise, you might have cases where the source link is not added.
- `m_menu` should be named `m_Menu` to follow conventions.

And then a last one that I am happy to overlook in this PR but need to be fixed eventually: It is technically possible for the source link to be outdated if the component was able to Switch between file names. It cannot at the moment but we might want to change that in the future to match the behaviour of the `Type` component. If you don't fix it now, make sure to raise an issue for it.

adecler (approved):
Now looking good after discussion and commit

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check required

FraserGreenroyd (approved):
I have tested this with a few datasets which included valid source links, invalid source links, and no source links. At all times, the behaviour was as expected based on the issue described and reviewed against.

This was previously approved for merge prior to the feature freeze implemented for the end of milestone. Although we are now in feature freeze, I am of the opinion that this addition to the UI code increases usability, user experience, and ability to question dataset sources to further increase confidence in datasets.

As such, in my capacity as DevOps lead, I am granting this PR permission to merge during feature freeze for deployment in the 4.3 beta.

PR #414, opened 2022-01-21
Add support for single matching input for Create method replacement of Type

IsakNaslundBh:
### NOTE: Depends on 

Done in support of https://github.com/BHoM/BHoM_Engine/pull/2738/files but does not require that to compile.
   
Adding support for single type matching of inputs to be replaced when going from Create method to PropertyAssignment create method. This is what is done on the updates side, so should be a simple alignment with what is done there.


IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check required

FraserGreenroyd (approved):
Have run through with @IsakNaslundBh and am happy the addition to the code base allows parameter matching by data type in the instance where the parameter name has changed. Works for the use case described.

PR #456, opened 2023-05-18
Utilise Exposure in ParamInfo

### LINKED ISSUE #455: Utilise UIExposure on method inputs
In line with https://github.com/BHoM/BHoM/pull/1506 and https://github.com/BHoM/BHoM_Engine/pull/3052 - include this exposure detail in deciding whether or not to display an input to a component.

FraserGreenroyd:
Fixes #455 

This provides the UIExposure information to the `ParamInfo` object used by all UIs downstream.

Each UI will be responsible for handling how they opt to display or not display something with `IsExposed = false` depending on each UIs requirements.

### DIFF
--- a/UI_oM/ParamInfo.cs
+++ b/UI_oM/ParamInfo.cs
@@ -51,6 +51,8 @@ namespace BH.oM.UI
 
         public virtual string DefaultValueWarning { get; set; } = "";
 
+        public virtual bool IsExposed { get; set; } = true;
+
         /***************************************************/
     }
 }
--- a/UI_oM/ParamInfo.cs
+++ b/UI_oM/ParamInfo.cs
@@ -51,8 +51,6 @@ namespace BH.oM.UI
 
         public virtual string DefaultValueWarning { get; set; } = "";
 
-        public virtual bool IsExposed { get; set; } = true;
-
         /***************************************************/
     }
 }
--- a/BHoM_UI/Caller/SetInputs.cs
+++ b/BHoM_UI/Caller/SetInputs.cs
@@ -67,7 +67,7 @@ namespace BH.UI.Base
                     ParamInfo p = Engine.UI.Create.ParamInfo(x, descriptions.ContainsKey(x.Name) ? descriptions[x.Name] : "");
 
                     if (exposures.ContainsKey(x.Name) && exposures[x.Name] == UIExposure.Hidden)
-                        p.IsExposed = false;
+                        p.IsSelected = false;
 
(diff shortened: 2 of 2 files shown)


FraserGreenroyd:
@BHoMBot check versioning

UI_oM/ParamInfo.cs
```diff
@@ -51,6 +51,8 @@ public class ParamInfo : BHoMObject
 
         public virtual string DefaultValueWarning { get; set; } = "";
 
+        public virtual bool IsExposed { get; set; } = true;
```
IsakNaslundBh, line 54: I think there is overlap between this new property and the `IsSelected` property, which essentionally does the same thing, at least on the level of Grasshopper (IsSelected == false -> input not on the component).

Might be better to try to align the two in this case, also to avoid having to re-implement the running mechanisms, fetching default values, etc.

Basically, a bit scared of the overlap, and a simple solution might just be to remove this property and simply set the IsSelected instead, which might just finish the job without the need for modification in any other part of the code, like for example GH.

FraserGreenroyd:
@BHoMBot check core
@BHoMBot check versioning
@BHoMBot check installer

UI_oM/ParamInfo.cs
FraserGreenroyd: Have done - per our offline discussion, we're happy for advanced/super users to allow `hidden` inputs to be visible in the philosophy of help (by hiding) without constraining (making inaccessible).

IsakNaslundBh (approved):
Tested in grasshopper with example method.

```
        [Description("Queries the centre of area enclosed by a closed, planar, non-self-intersecting Polyline.")]
        [Input("curve", "The Polyline to get the centre of area of.")]
        [Input("tolerance", "Distance tolerance used in geometry processing, default set to BH.oM.Geometry.Tolerance.Distance.", typeof(Length))]
        [Input("validateInput", "Toggels whether checks on the input geometry should be run or not, defaulting to true.", oM.Base.Attributes.Enums.UIExposure.Hidden)]
        [Output("centroid", "The Point at the centre of a region enclosed by given Polyline.")]
        public static Point Centroid(this IPolyline curve, double tolerance = Tolerance.Distance, bool validateInput = true)
        {
...

```

And functions as intended.

Only comment, as discussed offline, is regarding how the method shows up in menu-selection, both global search and right-click menu on component. As discussed this will be tackled by https://github.com/BHoM/BHoM_Engine/issues/3053

Code changes also are small and makes sense.

With all this, happy to approve.

PR #479, opened 2024-01-04
Solving the slow execution of CollectOutputTypes on Custom objects

### LINKED ISSUE #478: Explode of custom objects take really long time to delete
Putting this as a question/bug, as not sure what is going on..

The explode component takes a really long time to delete from the GH canvas for custom objects if you have many of them.

Can understand that the creation of the explode might take some time, as it requires to check all the custom data etc, but do not fully understand why it takes long time to delete.

For a simple script like the below, it takes 16 seconds to just delete the explode (literally selecting it and hitting the delete key).

image

The custom objects are ofc really useful to help with grouping and filtering, and have recently been trying to do that with results, where you easily can end up with millions of datapoints. Have hade the script freezing for quite some time after I decided to delete some explode in some of those scripts. There is always a workaround by using GetProperty instead (which I have resorted to to avoid the issue), but good to understand what is going on

adecler:

Closes #478 

### Test files
[ExplodeDeleteTest.zip](https://github.com/BHoM/BHoM_UI/files/13833322/ExplodeDeleteTest.zip)

The call to `OutputParams` was very slow (~20s) so I added it to the scrip to confirm in now takes only a few milliseconds.

### DIFF
--- a/UI_Engine/Query/OutputParams.cs
+++ b/UI_Engine/Query/OutputParams.cs
@@ -148,7 +148,7 @@ namespace BH.Engine.UI
 
         private static void CollectOutputTypes(IEnumerable<CustomObject> objects, ref Dictionary<string, List<Type>> properties)
         {
-            foreach (KeyValuePair<string, object> prop in objects.SelectMany(x => x.CustomData).Distinct())
+            foreach (KeyValuePair<string, object> prop in objects.SelectMany(x => x.CustomData))
             {
                 if (!properties.ContainsKey(prop.Key))
                     properties[prop.Key] = new List<Type>();


FraserGreenroyd (approved):
Approving based on code review and discussion with @adecler during the historic issue closure today, this resolves the speed issue and happy to deploy to alpha to be tested by a broader user base during the milestone to check it isn't causing any problems with things we haven't considered today.

FraserGreenroyd:
@BHoMBot check versioning
@BHoMBot check installer
@BHoMBot check core

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #482, opened 2024-02-26
Introduce settings options within the UI framework

### LINKED ISSUE #215: BHoM_UI: tags and local settings
#### Description:

Following the discussion with @al-fisher and @kThorsager, I am writing down the ideas for UI improvements that we came up with:
- Method tags, possibly really helpful to identify methods that are e.g. related to a certain workflow. These could be divided into:
    - Default tags - either hard coded in the source or stored in a local file
    - Custom tags - to be stored locally in _%appdata%_

    Plus @al-fisher's idea: a ❤️ button to mark favourite methods
- A window with data tree, where the user could check/uncheck each assembly/method to include or remove it from the search (better UX plus faster search?)

To be compared against #85

Suggestions and thoughts are welcome!

FraserGreenroyd:
### NOTE: Depends on 
https://github.com/BHoM/BHoM_Engine/pull/3296 - this helps make this work for testing, but is not strictly a dependency given the outline of this work.

   
Fixes #215 

### Test files
There is an identically named branch on Grasshopper_UI to help test this.
 - On the UI tab of BHoM, there is a component called UI Settings (see image) - it shares an icon with the Cluster to Node for now because it is a testing component and not the finished product
 - Drop the component on the canvas, you should get the pop up to choose which toolkits to have in search results (see image)
 - Play around with that window and change items, save and reset, etc.
 - When saving you'll find the component disappeared. Redrop it to get the pop up back
 - Close and reopen Rhino to check that settings persist when saved/closed
 - You can check the `ctrl` `shift` `b` menu for items being excluded which is introduced in this PR, but is not intended to be available fully until a later milestone - but as proof of concept, my example was to untick `Acoustic`, `Analytical`,  and `Facades` and then search for `Panel` and the first `Panel` object to show up is the Environment Panel

image
image

### Changelog
 - Add project to handle UI Settings windows that are common between all UIs
 - Add UI window for handling turning on/off toolkits in search results

### Additional comments
This PR introduces the concept of settings windows for BHoM to provide greater flexibility in user set up and enhance UX across the BHoM UIs which work from this UI framework.

The work covered in this PR is NOT designed to be utilised by any UI in this milestone (7.1), as greater thought needs to be given to the individual settings each UI might want to make. For example, the ability to search when wires are dropped is a Grasshopper specific setting as Excel does not perform the same action. However, both UIs can benefit from only having certain items appear in their `ctrl` `shift` `b` menus based on user selection.

This PR will thus be excluded from versioning needs if changes need to be made for final implementations by others, but provides the starting blocks on which we can then aim to build. As such, perfection is not necessarily required by this PR providing it is not glaringly problematic with what's introduced, because this simply being here will not impact any user experience until explicitly called upon by a UI implementation.


FraserGreenroyd:

FraserGreenroyd:
@BHoMBot fix project file ref. 21976635563

FraserGreenroyd:

Tom-Kingstone (approved):
Testing with the branch in Grasshopper_UI, Works as expected - When deselecting toolkits like PowerPoint and then saving, their respective methods and objects are removed from the search menu. If I close and reopen rhino+gh, the settings are saved, and still cannot search the objects/methods deselected. Then when resetting and saving, the changes persist between instances of rhino as well. The toolkits are also omitted from the grasshopper drag/drop search.

Tom-Kingstone (approved):
Checked the new select/deselect all button, and it correctly selects and deselects all the inputs, and from a colourblind perspective the buttons are all distinct as well.

Tom-Kingstone:
@BHoMBot check core

Tom-Kingstone:

albinber (approved):
Select all feature looks great and improves UX. Everything else works nicely, happy to approve. Thanks @FraserGreenroyd :)

FraserGreenroyd:
@BHoMBot fix copyright headers ref. 21992462939

Tom-Kingstone (dismissed):
approving based off previous approval (only headers changed)

FraserGreenroyd:
@BHoMBot check versioning
@BHoMBot check installer
@BHoMBot check core

michaelhoehn (changes requested):
Nice! I really like this and it's a great first step to expanding and improving our UI UX. Looking forward to seeing this implemented in our supported UIs. 

I did test in grasshopper using the matching branch. I'll make some comments on an issue in GH UI to talk about specific implementations there. 

A few requested changes I'll note here: 

1. If I'm a user and I've updated my settings but in doing so, I've accidentally removed a relevant namespace, it would be good to prompt the user to modify the settings again if they are not seeing what is expected. For example: if noPossibleItems & user has filter on then tell user to check settings. This would be changing the UI message from "_No result found..._" to something like "_Try adjusting the UI search settings and try again_". 

2. The window seems a bit rigid and can't be moved or minimised. Would expect to be able to move/close/minimise.

FraserGreenroyd:
Thanks for the review @michaelhoehn .

For item 1, I think that will need to go in as a new feature/enhancement request because that will need to update across a few parts of the search (simply adding a check in the `GlobalSearch` isn't enough unfortunately as there are other places it updates the search results based on a search term).

I have implemented item 2 however if you wish to take another look 😄

FraserGreenroyd:
@BHoMBot check versioning
@BHoMBot check installer
@BHoMBot check core

michaelhoehn (approved):
Thanks @FraserGreenroyd 👍 
This looks really great. Excited for the direction.

pawelbaran:
Good stuff @FraserGreenroyd - I had not looked at the code but enjoyed the test in GH 👍  A few comments from my side:
- The choice of available checkboxes looks quite random: why did `AddinManagementSettings` get its dedicated item? There is quite of few of such surprises, definitely worth having a closer look at reflection mechanism, a bit of curation would not hurt either (this is actually a general comment, would be great to have a thorough go through what and where is reflected in the UI - I have seen a few idiosyncrasies in the past, pretty sure we managed to grow many of them over years)
- Save/load buttons would be desirable, especially as we go into more detail in the settings (see teams sharing common UX, especially valuable for newcomers)
- Discipline-specific presets would be a nice to have, I can imagine people scratching their heads under half of the items thinking whether it is relevant for their work
- Expert/programmer mode could also be helpful - most of people won't need namespaces such as `Versioning`, `CSharp` or `Attributes`
- I know it is a rough prototype, but the disappearing GH component looks awkward 😃 

...but of course this is all phase 2 I would say, this PR seems to be a good start!

FraserGreenroyd:
Thanks for the feedback @pawelbaran 😄 To answer some points below:

> * The choice of available checkboxes looks quite random: why did `AddinManagementSettings` get its dedicated item? There is quite of few of such surprises, definitely worth having a closer look at reflection mechanism, a bit of curation would not hurt either (this is actually a general comment, would be great to have a thorough go through what and where is reflected in the UI - I have seen a few idiosyncrasies in the past, pretty sure we managed to grow many of them over years)

This is a good question. The bit of code responsible for handling Toolkit items is:

```
var toolkits = fullNames.Select(x =>
{
    var split = x.Split('.');
    if (split.Length < 3)
        return null;

    var toolkit = split[2];
    if (toolkit == "Adapters" || toolkit == "Revit" && split.Length > 3)
        toolkit = split[3];

    return toolkit;
}).Where(x => !string.IsNullOrEmpty(x)).ToList();
```

Which is in the `LoadFromLoadedAssemblies()` method for the window. I have attempted to handle adapters namespace and the Revit namespace but thinking about it, maybe this is the issue? The Revit oM namespace is presumably `BH.oM.Adapters.Revit` whereas other Revit assemblies may be `BH.oM.Revit.AddinManagementSettings`?

I agree it's one to take another look at before we turn it on fully in any of the UIs though - a workshop with a few of us to bottom out and check objects are suitable with namespaces should suffice - I'll look to schedule something in soon 😄 

> * Save/load buttons would be desirable, especially as we go into more detail in the settings (see teams sharing common UX, especially valuable for newcomers)

The settings themselves load/save using the [Settings_Engine](https://github.com/BHoM/BHoM_Engine/tree/develop/Settings_Engine) which will automatically save any settings to the `%ProgramData%/BHoM/Settings` folder and also loaded from there on start up. Thus, sharing a setup could be done by anyone sharing the JSON settings file from that folder I think?

The Settings_Engine does need documenting better (on BHoM.xyz) to be fair, which is slated for the end of this milestone which may benefit. But equally happy to workshop around additional save/loading options - perhaps a load from file for an initial set up for people who may not be comfortable pasting the file into `%ProgramData%/BHoM/Settings`?

> * Discipline-specific presets would be a nice to have, I can imagine people scratching their heads under half of the items thinking whether it is relevant for their work

Agreed - I'm hoping the discipline communities can help with this 😄 

> * Expert/programmer mode could also be helpful - most of people won't need namespaces such as `Versioning`, `CSharp` or `Attributes`

Agreed, and I think falls under point 1 above as well of checking what some names are (`Attributes` shouldn't be its own toolkit for example) and possible drawing up a categorisation system as well for Toolkits would be beneficial (long term strategy I think).

> * I know it is a rough prototype, but the disappearing GH component looks awkward 😃

Also agreed, which is primarily why I opted to do this PR first without doing a conjoined PR on Grasshopper, I wasn't comfortable with the approach based on the lack of buttons on ribbon option I could find so want to give that more thought/input before it becomes available. I'd also like (though it may not be feasible given resource available) to do both GH and Excel at the same time, and also, liaise with you to do Revit, so all 3 UIs can benefit potentially at once - but that needs more mapping out and workshops I think for planning week for coordination as well 😄 

That said, @michaelhoehn had a reasonable idea and has documented it [in this issue](https://github.com/BHoM/Grasshopper_UI/issues/698) - I think it's worth crowding round that issue for the Grasshopper implementation specifically (and eventually the Excel one, and eventually the Revit one, and eventually the new UI one, etc., etc.).

pawelbaran:
Thanks @FraserGreenroyd, all makes sense, just one immediate thought on your replies:

> The settings themselves load/save using the [Settings_Engine](https://github.com/BHoM/BHoM_Engine/tree/develop/Settings_Engine) which will automatically save any settings to the `%ProgramData%/BHoM/Settings` folder and also loaded from there on start up. Thus, sharing a setup could be done by anyone sharing the JSON settings file from that folder I think?

What we as programmers find easy, for others may be uncomfortable or even worse, a no-go. Digging in 'system' files scares off many individuals from my experience, while everyone is fine with clicking 'Save' in the UI. So I would still opt for a button, as long as the users do not disagree 😉

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #511, opened 2025-03-20
Add ProjectId to triggerUseageLogArgs and check if set and update ProjectId

IsakNaslundBh:

Add support for retreiving projectId from events fired. If projectId is set upon return from TriggerUseageLog, then Id is used for the current file and set to update and stored for future useage.


michaelhoehn (approved):
I've tested with multiple project and non-project files in both Excel and Rhino 7 and have found this solution functional. Happy to approve.

IsakNaslundBh:
@bhombot check copyright-compliance
@BHoMBot check ready-to-merge

PR #517, opened 2025-04-15
Add support for dynamic objects

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/1677

   
Addresses the issue raised in the BHoM. See https://github.com/BHoM/BHoM/pull/1677 for more details.

### DIFF
--- a/BHoM_UI/Caller/Compile.cs
+++ b/BHoM_UI/Caller/Compile.cs
@@ -69,7 +69,7 @@ namespace BH.UI.Base
             for (int index = 0; index < OutputParams.Count; index++)
             {
                 ParamInfo param = OutputParams[index];
-                Func<IDataAccessor, object, int, bool> function = Engine.UI.Compute.OutputAccessor(accessorType, param.DataType);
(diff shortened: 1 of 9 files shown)


adecler:
@BHoMBot check compliance

IsakNaslundBh (commented):
Generally works well. Only issue left to handle is creation of subclass of Quantity<T> in the UI does not work, as it gets turned into a numeric type for the output accessor. Might be that we simply exclude them from ever showing up in the UI, or alternatively, make an exception for create methods, to let them actually output the class, rather than double with quantity attribute on the description.

IsakNaslundBh:
Another nitty gritty thing found. Would be great if the dynamic properties could be sorted on the explode, preferably by the underlying enum value:
image

IsakNaslundBh:
@BHoMBot check versioning

IsakNaslundBh (approved):
Now working as expected in all cases I have tested for. Code changes also make sense.

Issues found with order fixed:
image

Also now able to create Quantities, while they still show up as numbers on outputs.

Happy with all the changes in here to be merged. Thanks @adecler

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #523, opened 2025-05-23
Group events and set max events to show in log

### LINKED ISSUE #522: Versioning log takes to long to render for an extreme amount of items
#### Description:

The versioning log window can take an quite extreme time to generate when a large set of versioning events has been thrown. This can happen if you open a script with data/datasets requiring to upgrade, and hence being put in the versioning log. The current solution generates a table for every single item being upgraded.

This is even more pressing to fix after https://github.com/BHoM/BHoM_UI/pull/509 where the log was moved to the main UI thread.

The time it takes to generate as well as draw the form seems to slow down exponentially with the number of items drawn, and after a couple of 10's it becomes far to slow. And with slow we are talking minutes, not seconds.

To solve, suggestion two things:
1. Group all events that display the same message anyway, and add a counter to the empty top box of the table
2. Limit the number of events shown. Suggesting maximum number of 30 (half arbitrarily picked from testing)

Slightly more long term we should aim to get this window back onto its own thread again, but then need to find a a good replacement for the discontinued thread.Abort().

#### Steps to reproduce:

Open a script with a multitude of objects having to be upgraded

#### Expected behaviour:

For the UI to not freeze on startup

IsakNaslundBh:

Closes #522

See #522 for more detials.

Fixes problem with extreme load time for the versioning log.

### Test files

Any file that requires versioning

### DIFF
--- a/BHoM_UI/Global/DocumentListener.cs
+++ b/BHoM_UI/Global/DocumentListener.cs
@@ -84,6 +84,7 @@ namespace BH.UI.Base.Global
                 }
 
                 ShowForm(events);
+                //Task.Run(() => ShowForm(events));
             }
         }
 
@@ -115,10 +116,6 @@ namespace BH.UI.Base.Global
             if (events == null)
                 return;
 
-            string message = $"The file was upgraded from BHoM version {events.First().OldVersion} to version {events.First().NewVersion}.";
-            message += "\nPlease review the file before saving it.";
-            message += "\n\nHere's the list of components that have been modified:";
-
             m_VersioningForm = new Form
             {
                 Text = "BHoM versioning report",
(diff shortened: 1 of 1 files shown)


michaelhoehn (dismissed):
I've tested in both Rhino 7 and 8 with success. Happy to see this merged however I would agree with Arnaud's comments above on limiting. Although I'm not opposed to imposing a fixed display limit personally (since 30 seems like a sufficient amount), I would recommend another feature be added later on to address the creation of a log and linking to that. I think the scope of this PR is sufficiently accomplishing it's intended goal.

michaelhoehn:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@michaelhoehn thanks for the review. I had already almost finished off the versioning log file, so thought I might as well get it in. Please are you able to re-review. To test the look of the log file, you can set the `maxMessages` to 1 temporarily.

michaelhoehn (approved):
Happy with the functionality. I have tested log generation successfully as well.

michaelhoehn:
@BHoMBot check ready-to-merge

PR #531, opened 2025-10-19
Avoid loading all assemblies by using a pre-compiled menu data file

### LINKED ISSUE #521: Speed up the loading time of the BHoM UI
#### Description:
Needs to be working both in Excel and Grasshopper

adecler:
### NOTE: Used by 
https://github.com/BHoM/Excel_UI/pull/423

### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/1698
https://github.com/BHoM/BHoM_Engine/pull/3561
https://github.com/BHoM/BHoMAnalytics_Toolkit/pull/95
https://github.com/BuroHappoldEngineering/BuroHappold_BHoMAnalytics_Toolkit/pull/82
https://github.com/BuroHappoldEngineering/BuroHappold_Installer/pull/462
https://github.com/BHoM/Versioning_Toolkit/pull/317
   
Closes #521 

This PR eliminates the need to load all the assemblies when starting the BHoM UI. Instead, here's how it now works:
 - Menus are populated using a pre-compiled file containing all the items found in the menus. 
 - Those menu items are first created with just the json representation of the method/object it links to. 
 - When a menu item is selected for the first time, the json will be deserialised to link the method/object to it.
 - The serialiser itself in in charge of dynamically loading the dlls needed to deserialise an object. Those that haven't been loaded already at least. 

Solving the problem at the level of the serialiser means we are taking care of the following problems at the same time:
 - UI component creation
 - UI component copy-pasting
 - File saving
 - Toolkit setting
 - Deserialisation of BHoM datasets

The file that contains all the information needed to populate the menus is located here: `C:\ProgramData\BHoM\Resources\AssemblyContent.tsv`

### Test files
In both Excel, Rhino 7, and Rhino 8:
  - Try to create each type of component from the global menu. Make sure each component links to a method/object from a different. assembly.
  - Try to create each type of component from the components found in the ribbon. Make sure each component links to a method/object from a different. assembly.
  - Save the file, close BHoM_UI and re-open the file to make sure everything was saved correctly
  - Deserialise BHoM data from existing datasets. Make sure to test versioning by using older datasets
  - Check that the analytics is still working

While testing you can use the "AllEvents" components to check the initialisation log. It contains quite a bit of details on what was happening and how long it took: 
image

### Notes on assemblies loaded at each start
If you check the event log, you'll notice that there are a set of assemblies that are loaded because they were identified as new, even if the AssemblyContent.tsv file is up to date with the dlls. This is because those dlls are empty and should be removed from their respective repo:
- **AGS_Engine**
- **Buildings_Engine**
- **Clipper_oM**
- **Radiance_Adapter**: Not fully empty but I Adapter is marked as to be removed
- **RFEM5_Engine**
- **RFEM6_Engine**

There are a couple that will need further investigation though:
- **~~ETABS_Adapter~~**: There was a case mismatch between the assembly name and the file name. Fixed in this commit:  https://github.com/BHoM/BHoM_UI/pull/531/commits/e0deef5be3bd4e9a8141d3776cc981f4a647dc93
- **IES_Engine**: Just has a python script, not sure what to do with this one
- **Karamba3D_Engine**: Fails to load assembly. Exception message: Could not load file or assembly 'KarambaCommon, Version=3.0.0.4, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.

### Notes on starting time still sometimes being long
Before the code code for the BHoM_UI is even entered, Windows is scanning the entire Assembly folder and process each of them (without loading, just getting some metadata). This takes over a minute and is completely unnecessary. Unfortunately, this cannot be avoided and can only be solved by reorganizing the dlls into sub-folders. That will be a difficult task as dlls are referenced directly in each toolkit. So we will need extensive planning to solve this. Here's an analysis of the situation if you want a bit more details: 
[LoadingTimeAnalysis.md](https://github.com/user-attachments/files/25163149/LoadingTimeAnalysis.md)

### DIFF
--- a/BHoM_UI/Global/Initialisation.cs
+++ b/BHoM_UI/Global/Initialisation.cs
@@ -370,7 +370,7 @@ namespace BH.UI.Base.Global
         private static bool UpdateCodeElements(List<string> loadedAssemblies)
         {
             List<CodeElementRecord> loadedCodeElements = BH.Engine.UI.Query.CodeElements()
-                .Where(x => loadedAssemblies.Contains(x.AssemblyName))
+                .Where(x => loadedAssemblies.Contains(x.AssemblyName, StringComparer.OrdinalIgnoreCase))
                 .ToList();
 
             if (loadedCodeElements.Count == 0)
--- a/UI_Engine/Compute/LoadNewAssemblies.cs
+++ b/UI_Engine/Compute/LoadNewAssemblies.cs
@@ -53,6 +53,9 @@ namespace BH.Engine.UI
                 return new List<string>();
             }
 
+            // Make sure the keys for the assemblies are in lower case to avoid casing mismatching
+            Dictionary<string, DateTime> lastUpdateTimes = lastAssemblyUpdateTimes.ToDictionary(x => x.Key.ToLower(), x => x.Value);
+
             List<string> loadedAssemblies = new List<string>();
 
             Regex regex = new Regex(@"oM$|_Engine$|_Adapter$");
@@ -61,7 +64,8 @@ namespace BH.Engine.UI
                 string name = Path.GetFileNameWithoutExtension(file);
                 if (regex.IsMatch(name))
                 {
-                    if (!lastAssemblyUpdateTimes.ContainsKey(name) || lastAssemblyUpdateTimes[name] < File.GetLastWriteTimeUtc(file))
+                    string key = name.ToLower();
+                    if (!lastUpdateTimes.ContainsKey(key) || lastUpdateTimes[key] < File.GetLastWriteTimeUtc(file))
                     {
                         Assembly assembly = BH.Engine.Base.Compute.LoadAssembly(file);
                         if (assembly != null)


adecler:
@BHoMBot check versioning

BHoM_UI/Global/AssemblyResolver.cs
```diff
@@ -0,0 +1,110 @@
...
+        /**** Private Methods             ****/
+        /*************************************/
+
+
+
+        /*************************************/
+        /**** Private Fields              ****/
+        /*************************************/
+
+        Dictionary<string, List<string>> m_AssemblyNamePerType = new Dictionary<string, List<string>>();
```
pawelbaran: I would consider changing the type here to an array for performance purposes, assuming these are closed sets that do not change dynamically (and knowing there might be thousands of them)
```suggestion
        Dictionary<string, string[]> m_AssemblyNamePerType = new Dictionary<string, string[]>();
```

BHoM_UI/Global/Initialisation.cs
```diff
@@ -21,21 +21,27 @@
...
 using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
+using System.Windows.Forms;
```
pawelbaran: 😉 
image

UI_Engine/Compute/LoadNewAssemblies.cs
```diff
@@ -0,0 +1,82 @@
...
+        /*************************************/
+
+        [Description("Loads all BHoM assemblies from the current domain.")]
+        [Input("lastAssemblyUpdateTimes", "records of the last time each assembly was updated.")]
+        [Output("loadedAssemblies", "Assemblies loaded as considered new")]
+        public static List<string> LoadNewAssemblies(Dictionary<string, DateTime> lastAssemblyUpdateTimes)
+        {
+            List<string> loadedAssemblies = new List<string>();
+
+            Regex regex = new Regex(@"oM$|_Engine$|_Adapter$");
```
pawelbaran: Not a big fan of this one, it introduces misalignment with BHoM_Engine

UI_Engine/Compute/LoadNewAssemblies.cs
```diff
@@ -0,0 +1,82 @@
...
+            foreach (string file in Directory.GetFiles(BH.Engine.Base.Query.BHoMFolder(), "*.dll", SearchOption.TopDirectoryOnly))
+            {
+                string name = Path.GetFileNameWithoutExtension(file);
+                if (regex.IsMatch(name))
```
pawelbaran: Following my previous comment, I would suggest using the existing methods to centralise the core classification of assemblies
```suggestion
                if (name.IsOmAssembly() || name.IsEngineAssembly() || name.IsAdapterAssembly())
```

UI_Engine/Query/Items.cs
```diff
@@ -136,8 +137,8 @@ public static IEnumerable<Type> ConstructableRequestItems()
         [Output("items", "All types valid in BHoM.")]
         public static IEnumerable<Type> TypeItems()
         {
-            return Engine.Base.Query.BHoMTypeList()
-                .Concat(Engine.Base.Query.BHoMInterfaceTypeList())
+            return Engine.Base.Query.AllTypeList()
```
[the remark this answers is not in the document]
adecler: Limiting the types registered exclusively to the BHoM objects causes multiple failure on serialisation. 
We have separate items for specifically creation BHoM objects via their set of properties or `Create` methods. So the Types registered here are purely for deserialisation and creating `Type` objects in the UI.

michaelhoehn:
@BHoMBot this is a DevOps instruction. Run a force check on: installer

michaelhoehn (approved):
I have tested the code in Rhino 7, 8, and Excel (using https://github.com/BHoM/Excel_UI/pull/423) with satisfactory testing results. 

Testing methodology: 
- Test only this PR against BHoM_Engine `develop` branch (all UIs confirmed)
- Test this PR against BHoM_Engine `BHoM_Engine-#3567-ASsemblyResolverOnExtensionMethods` branch (all UIs)
- Each test verified the initialisation and creation of all UI components from the global menu as well as spot checks for a wide variety of namespace methods/objects.
- confirmed Rhino version compatibility
- confirmed usage analytics functionality

Personally happy with the speed and results I'm seeing across the UIs. Good from my perspective to merge.

PR #546, opened 2026-05-03
Fix bug on null outputs for the Explode component

### LINKED ISSUE #499: Explode fails to return all values when one entry is null
#### Description:
See this example: 
image

#### Steps to reproduce:
Use this script: 
[ExplodeBug.zip](https://github.com/user-attachments/files/17666855/ExplodeBug.zip)

#### Expected behaviour:
All data should be returned from the explode regarding of the value found on other object in the input list

adecler:

Closes #499

See issue for details.

The root of the issue is that the explode component is trying to assign null to a type that is not nullable. In the case of the example script the type is double. This throws an exception that is caught but thrown again so it breaks the outputs for the following outputs of that connector. 

### Test files
You can test with this file:
[ExplodeBug.zip](https://github.com/user-attachments/files/27321030/ExplodeBug.zip)

### Additional comments
**Important!!!** 
You need to recompile Grasshopper after compiling BHoM_UI for the dlls to be copied to the Grasshopper folder in C:\ProgramData\BHoM\GrasshopperPlugin

### DIFF
--- a/BHoM_UI/Caller/Run.cs
+++ b/BHoM_UI/Caller/Run.cs
@@ -204,6 +204,20 @@ namespace BH.UI.Base
                     {
                         if (OutputParams[i].IsSelected)
                         {
+                            if (output == null)
+                            {
+                                var subType = OutputParams[i].DataType.UnderlyingType();
+                                Type elementType = subType.Type;
+                                if (QuantitiesAsDouble && typeof(BH.oM.Quantities.IQuantity).IsAssignableFrom(elementType))
+                                    elementType = typeof(double);
+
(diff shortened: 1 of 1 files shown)


michaelhoehn (approved):
This appears to function as expected so I'm happy to proceed with merge. I will leave this test result here for quick future discussion as it still may seem a bit confusing based on the outputs. 

As you can see the `GetProperty` still returns a few nulls. This is because this property simply doesn't exist on the original json object for the first three indicies. So the behaviour here is correct and a null should be returned from the `GetProperty`.

Image

The major improvement here is the behaviour of the explode component as seen in the example taken from the original issue itself below.

Image

adecler:
@BHoMBot check ready-to-merge

PR #548, opened 2026-05-07
Fix filtering of the wire menu

adecler:

Closes #547

As described in the issue, the wire menu (created when dragging from a component connector in GH) was not provided useful element most of the time. This was due to to a change in our assembly loading strategy. Since the assemblies are now loaded dynamically, all the information needed to properly populate that menu was not available anymore. 

The strategy to solve has been to add two additional columns in the AssemblyContent.tsv file with the information necessary to due that filtering correctly (i.e. providing input and output types of each component alongside their base types and interfaces).

### Test files
Play with the wire menu as much as you can on different components, different types of expected inputs and outputs and make sure that the suggestions from the wire menu are making sense. Don't hesitate to comment if the order of those suggestions also seems off to you. 

**Important**: 
 - Make sure to recompile Grasshopper after compiling the BHoM_UI otherwise you will not see the effect in GH.
 - make sure to delete the AssemblyContent file in C:\ProgramData\BHoM\Resources. We need to generate a new one for those additional columns to be created. This is only a problem for testers compiling the code themselves, the installer would clean the file anyways.

### DIFF
--- a/BHoM_UI/Global/GlobalSearch.cs
+++ b/BHoM_UI/Global/GlobalSearch.cs
@@ -160,8 +160,12 @@ namespace BH.UI.Base.Global
             if (settings != null)
                 excludedToolkits = settings.Toolkits.Where(x => !x.Include).Select(x => x.Toolkit).ToList();
 
-            if (config != null)
+            if (config?.TypeConstraint != null)
             {
+                Type constraint = config.TypeConstraint;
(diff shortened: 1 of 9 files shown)


michaelhoehn (approved):
Tested successfully and approved for merge. I felt the wire menu felt a lot more natural and functional so big improvement 👍

PR #550, opened 2026-05-19
Adding runtime-aware assembly loading strategy

adecler:

Closes #549 

With this PR, it is now possible for individual toolkits to compile for multiple target frameworks. 
- By default, dlls should still go the the root Assemblies folder when they are compatible with all frameworks (i.e. running correctly in all BHoM  UIs)
- All dlls specific to .NET framework (i.e., not working in a .NET Core framework) should go in the `netfx` sub-folder
- All dlls only compatible with .net core frameworks, should go in the sub-folder corresponding to that framework's name. For example dlls for `.net7.0` should go in the `net7.0` sub-folder.  

### Test files
The best way to test this at the moment is to try the PR for the SQL toolkit: https://github.com/BHoM/SQL_Toolkit/pull/82
This PR is using the new system from this PR to solve the current incompatibility of the SQL adapter with Rhino 8

### Additional comments
**!!Important!!**: As always don't forget to compile Grasshopper UI after BHoM_UI for the changes to take effect.

### DIFF
--- a/BHoM_UI/Global/AssemblyResolver.cs
+++ b/BHoM_UI/Global/AssemblyResolver.cs
@@ -84,7 +84,7 @@ namespace BH.UI.Base.Global
             {
                 if (!BH.Engine.Base.Query.IsAssemblyLoaded(assemblyName))
                 {
-                    string assemblyPath = Path.Combine(BH.Engine.Base.Query.BHoMFolder(), assemblyName + ".dll");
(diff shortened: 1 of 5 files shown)


michaelhoehn (approved):
I have successfully tested compilation of the code using the SQL toolkit PR referenced above as well as UI functionality in both Rhino 7 and 8. The code functions as expected with no loss of funcitonality of other toolkits. 
Image

PR #553, opened 2026-06-11
Improve the UI Search settings

### LINKED ISSUE #497: Global search setting filters to broadly
#### Description:

Search filtering introduced in https://github.com/BHoM/BHoM_UI/pull/482 was a great addition to help simplify the UI to make it only show methods and classes relevant to you. However, the filtering mechanism introduced here:

https://github.com/BHoM/BHoM_UI/blob/a5e78cd8db0a39e9fcfea54b385631d96a78e986/BHoM_UI/Global/GlobalSearch.cs#L165-L186

Is to broad, and can filter out things not expected to be missing. For example, filtering out "Environment" makes Most LifeCycleAssessment methods not show, given they contain classes that contain "Environment" in their name.

#### Steps to reproduce:

Untick "Environment" and try searching for LCA_Toolkit objects and methods.

#### Expected behaviour:

Would expect the filtering mechanisms to filter based on namespace rather than the word being any part of the name. Suggesting to change the current filtering mechanisms that is a "Contains" to instead exclude methods and classes which types begins with for example, `BH.oM.Environment` and `BH.Engine.Environment` rather that the current implementation with a "Contains"

#### Test file(s):

Not really applicable, tested with settings and search menu behaviour.

adecler:
Closes #497 

This fixes the issue highlighted in #497, that search item were excluded just because the word for an excluded toolkit was appearing somewhere in their textual representation. This PR fixes that by looking specifically at the part of the namespace that relates to the toolkit when excluding search items.

This PR also fixes a series of other issues discovered when working on this PR:
 - The search setting file was re-opened every time a search was happening. This was highly inefficient so the filtering now happens a single time at start when loading the search item from the AssemblyContent file.
 - This also fixed the fact that only the global menu was filtering out the excluded toolkits. Components menus were still showing the excluded toolkits.
 - The search settings had their own UI window. This is different from any other settings in the BHoM since all other settings can be created using classic BHoM component. So I have removed the whole UI project but kept the bespoke access of the component from the ribbon to avoid confusion. This creates duplication with the fact that the settings can also be created from the CreateObject component but it is more user-friendly. We could also consider in the future to repurpose that component to create any settings so it makes sense to keep it for now. 

### Test files
 - Just create a search setting using this script:
image

 - You can check that the file was correctly created by looking at `C:\ProgramData\BHoM\Settings\BH.oM.UI.SearchSettings.json`
 - You can then restart the UI (close Rhino) and check that the toolkits are now excluded from the search menus. Make sure to test both the global menu and the component menus.


adecler:
@BHoMBot check required

michaelhoehn (approved):
Tested and approved. Happy with the functionality.

